PR #15, opened 2019-09-13
set FNMSKH=${FNMSKH:-${FIXam}/global_slmask.t1534.3072.1536.grb}

yangfanglin:
instead of
FNMSKH=${FNMSKH:-${FIXam}/seaice_newland.grb}
to use high-res land-sea mask in global_chgres.sh and global_cycle.sh for interpolation of climatological SST etc.

This is needed to reduce lake temperature biases for GFS.v16 development.


GeorgeGayno-NOAA:
One more change is needed.  Will reject this pull request.

PR #43, opened 2020-01-11
Catch unimplemented build method, add support for macOS compilers

climbfuji:
This PR adds a guard inCMakeLists.txt to prevent building UFS_UTILS using cmake standalone. This has not been implemented yet, because we are now using the findESMF macro directly from ESMF. Using this macro requires a bit of additional logic (see NCEPLIBS umbrella build CMakeLists.txt, branhf ufs_release_v1.0).

This PR also adds support for macOS with clang compilers, which somehow slipped in previous PRs. I am going to merge this PR, because it doesn't remove/change any working logic.

Note: the following CMakeLists.txt files only know about the Intel compiler and return an "unknown compiler" warning for any other compiler. This needs to be fixed (see issue https://github.com/NOAA-EMC/UFS_UTILS/issues/44).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ endif()
 if(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
   message("setting intel true")
   set(IntelComp true )
-elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU*")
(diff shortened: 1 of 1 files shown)


climbfuji:
@DusanJovic-NOAA @mark-a-potts @kgerheiser for your information

Issue #61, opened 2020-02-19
release/public-v1: Fix legacy Fortran extension


DusanJovic-NOAA:
```
/home/dusan/simple-ufs/src/preproc/sorc/chgres_cube.fd/write_data.F90:1167:31:

 1167 |    WRITE(OUTFILE, '(A, I1, A)'), 'out.atm.tile', tile, '.nc'
      |                               1
Warning: Legacy Extension: Comma before i/o item list at (1)
/home/dusan/simple-ufs/src/preproc/sorc/chgres_cube.fd/write_data.F90:1637:33:

 1637 |      WRITE(OUTFILE, '(A, I1, A)'), 'out.sfc.tile', tile, '.nc'
      |                                 1
Warning: Legacy Extension: Comma before i/o item list at (1)
[100%] Linking Fortran executable chgres_cube.exe
```

GeorgeGayno-NOAA:
Dusan, I can add this small change to one of my other issues and then merge to 'develop'.  To get this in the public release branch, you will need to contact the branch code manager.

GeorgeGayno-NOAA:
What compiler is giving this warning?

climbfuji:
GNU does.

DusanJovic-NOAA:
GNU (gfortran)

> What compiler is giving this warning?

GNU (gfortran)

GeorgeGayno-NOAA:
Merged to 'develop' at 72c636c.  For convenience, I added this small fix to issue #64.

DusanJovic-NOAA:
Who is the code manager of release/public-v1 branch?

Issue #69, opened 2020-02-24
release/public-v1: syntax errors when compiled using debug flags (Intel)


DusanJovic-NOAA:
When I compile chgres_cube using debug compile flags (on Hera using Intel compiler), I get the following syntax errors:

```
/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/sufs/simple-ufs-aero/src/preproc/sorc/chgres_cube.fd/program_setup.f90(280): error #6631: A non-optional actual argument must be present when invoking a procedure with an explicit interface.   [RC]
     call error_handler("FOR GRIB2 DATA, PLEASE PROVIDE GRIB2_FILE_INPUT_GRID")  
----------^                                                                      
compilation aborted for /scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/sufs/simple-ufs-aero/src/preproc/sorc/chgres_cube.fd/program_setup.f90 (code 1)
```
and
```
/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/sufs/simple-ufs-aero/src/preproc/sorc/chgres_cube.fd/input_data.F90(5532): error #6631: A non-optional actual argument must be present when invoking a procedure with an explic
it interface.   [RC]
    call error_handler("ERROR USING MISSING_VAR_METHOD. PLEASE SET VALUES IN" // &
---------^
/scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/sufs/simple-ufs-aero/src/preproc/sorc/chgres_cube.fd/input_data.F90(5564): error #6631: A non-optional actual argument must be present when invoking a procedure with an explic
it interface.   [RC]
    call error_handler("reading soil levels. File must have 4 soil levels.")
---------^
compilation aborted for /scratch2/NCEPDEV/fv3-cam/Dusan.Jovic/sufs/simple-ufs-aero/src/preproc/sorc/chgres_cube.fd/input_data.F90 (code 1)
```

GeorgeGayno-NOAA:
Dusan, thanks for catching this.  The 'error_handler' routine could be better written.  It could be placed in a module.  And the return code could be an optional argument.

DusanJovic-NOAA:
Is there an easy way of fixing this in the public release branch?

GeorgeGayno-NOAA:
> 
> 
> Is there an easy way of fixing this in the public release branch?

Pass in a non-zero return code as the second argument.

[...]

PR #73, opened 2020-02-25
Fix error_handler call statements. Last argument (rc) must be present.

DusanJovic-NOAA:
See issue #69


DusanJovic-NOAA:
@GeorgeGayno-NOAA I also fixed issue #61

PR #75, opened 2020-02-25
Update submodule pointer for cmake 2020/02/25 - 2

climbfuji:
Update submodule pointer for cmake following the merge of https://github.com/NOAA-EMC/CMakeModules/pull/25. Can be merged immediately.


climbfuji:
Thanks George. The cmake submodule contains all the find_package macros for all NCEPLIBS, and I am just trying to keep the submodules updated (even if UFS_UTILS doesn't use it).

PR #76, opened 2020-02-26
Release/public v1

valbonakunkel-noaa:
in the docs folder, "CHGRES_CUBE.rst" file is added


valbonakunkel-noaa:
In my computer looks ok. I think maybe is gitlab format. I will ask Ligia
Laura about this and get back to you.

Valbona

PR #78, opened 2020-02-28
GDAS initializations scripts

GeorgeGayno-NOAA:
Create a set of utility scripts so a user can easily coldstart a cycled GDAS experiment. The scripts pull the needed data from HPSS and run the chgres_cube program for the high-res and ENKF members.

chgres_cube updates - Fix argument mismatches for three calls to the error handler.  Point to official v8 ESMF library on Dell and Cray instead of my own copy.  Point to my own copy of the wgrib2 api on Cray instead of Dusan's.


modulefiles/chgres_cube.wcoss_cray
```diff
@@ -7,20 +7,23 @@ module rm intel
...
+module use /gpfs/hps3/emc/nems/noscrub/emc.nemspara/soft/modulefiles
+module load esmf/8.0.0
+module rm gcc
+module load gcc/4.9.2
 
-export WGRIB2API_INC=/gpfs/hps3/emc/meso/save/Dusan.Jovic/wgrib2/include
-export WGRIB2_LIB=/gpfs/hps3/emc/meso/save/Dusan.Jovic/wgrib2/lib/libwgrib2.a
+
+export WGRIB2API_INC=/gpfs/hps3/emc/global/noscrub/George.Gayno/wgrib2/include
+export WGRIB2_LIB=/gpfs/hps3/emc/global/noscrub/George.Gayno/wgrib2/lib/libwgrib2.a
```
KateFriedman-NOAA: Will these be more officially located at some point (be in a non-personal space)?

GeorgeGayno-NOAA: And on Dell, I point to Wesley's home directory for this library.

I will keep pushing the NCEPLIBS team to formally implement it.  However, since chgres_cube is not used in operations yet, pointing to personal space should be ok for now.

Unfortunately, the community partners who wanted grib2 capability in chgres_cube made the decision to use wgrib2 instead of g2lib.  I would have chosen the latter, which is more flexible and formally supported.

KateFriedman-NOAA: Gotcha, thanks!

PR #84, opened 2020-03-02
Doc only: change to lower case

ligiabernardet:
Change file name and file content from CHGRES_CUBE to chgres_cube. Restructure subsections so they are now sections.


ligiabernardet:
Your comments have been addressed in the modified doc.

Issue #59, opened 2020-02-18
CHGRES_CUBE - range check SST


GeorgeGayno-NOAA:
During testing for the public release, the model failed (in debug mode) when using coldstart files prepared from GFS GRIB2 files.   The failure - in the nsst_module - indicates a too cold SST.  Investigate.  

See this issue for more details: https://github.com/ufs-community/ufs-mrweather-app/issues/87

GeorgeGayno-NOAA:
Too cold (< 271.2K) SST values were noted when using v14, v15.1 and v15.2 GFS GRIB2 data.  Also, a sample file from one of Russ' v16 parallels also had cold values.  These cold values are the result of bi-linear interpolation from the gaussian master files to the lat/lon "pgb" files.  As a result, coastal points (even those identified as water by the mask) can be a mixture of land and non-land.  The 0.5-degree data file had, on average, 300 cold points.

GeorgeGayno-NOAA:
Tried to reproduce the model crash on Hera and Dell (https://github.com/ufs-community/ufs-mrweather-app/issues/87) with no luck.  I used the GFS.v16.0.0 tag of the model and compiled it in debug mode.  I ran it for a global C96 canned case.  Even if I intentionally add very cold points (200 K) to the initial conditions, it still runs.  Not sure why that is.  Nevertheless, this problem should be fixed.   Will ask Xu Li for assistance.

GeorgeGayno-NOAA:
After consulting with Jun, she recommended I try to reproduce the error with the public release branch.  After some trial and error, I was able to do so.  What I did:

- Checked out release/public-v1 on Hera.
- Ran the "fv3_ccpp_gfs_v15p2_debug" regression test.
- Swapped out the coldstart files from the regression test with a new set from a modified version of chgres_cube that intentionally added very cold (260K) SST points.
- Reran the regression test.

The model broke before the write of the 00 hour history file:

`131: Image              PC                         Routine            Line        Source`
`131: fv3.exe            0000000003C60748  nst_module_mp_coo         894  module_nst_model.f90`
`131: fv3.exe            00000000036CC978  sfc_nst_mp_sfc_ns         392  sfc_nst.f`

That is the same line of code as the Cheyenne test: https://github.com/ufs-community/ufs-mrweather-app/issues/87

GeorgeGayno-NOAA:
[...]

PR #89, opened 2020-03-12
CHGRES_CUBE: range check SST when using GRIB2

GeorgeGayno-NOAA:
The GFS GRIB2 pgb files can contain SSTs colder than 271.2 K (because they are interpolated from the master file using bilinear interpolation).  Add range checking to the routine that reads the GRIB2 files.  For more details see issue #59.


GeorgeGayno-NOAA:
Xu, have you had a chance to look at this?  No hurry.

GeorgeGayno-NOAA:
> The range check applied to the water temperature is OK to me.

Thanks for your review.

Issue #91, opened 2020-03-16
Add CMake build to UFS_UTILS


GeorgeGayno-NOAA:
v1 of the public release used CMake to build `chgres_cube`.   Use CMake to build all programs in the repository.  Remove the old build system.

GeorgeGayno-NOAA:
Added the required cmake files from the public release branch to the 'feature/cmake' branch at cb887b6.  The "build_chgres_cube.sh" was updated to invoke cmake from Dell.   To get the config step to work, I had to updated the Modules/FindWGRIB2.cmake (under the git@github.com:NOAA-EMC/CMakeModules.git repo) as follows:
```
--- a/Modules/FindWGRIB2.cmake
+++ b/Modules/FindWGRIB2.cmake
@@ -5,7 +5,7 @@ endif()

 find_path (WGRIB2_INCLUDES
   wgrib2api.mod
-  HINTS ${WGRIB2_ROOT}/include)
+  HINTS ${WGRIB2_ROOT}/include ${WGRIB2_ROOT}/lib)
```

When I invoke the build script, the configure step works, but the build step fails:

```
-- Configuring done
-- Generating done
-- Build files have been written to: /gpfs/dell2/emc/modeling/noscrub/George.Gayno/ufs_utils.git/UFS_UTILS/build
+ make
Scanning dependencies of target chgres_cube.exe
[  8%] Building Fortran object sorc/chgres_cube.fd/CMakeFiles/chgres_cube.exe.dir/program_setup.f90.o
/gpfs/dell2/emc/modeling/noscrub/George.Gayno/ufs_utils.git/UFS_UTILS/sorc/chgres_cube.fd/program_setup.f90(343): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [ESMF]
  use esmf
------^
```

The path to the esmf module directory is not being set.  Not sure why that is.

kgerheiser:
CMake finds ESMF through the environment variable `ESMFMKFILE`. Is that set? It throws a warning when running cmake if it's not.

GeorgeGayno-NOAA:
Update the branch following Kyle's suggestion (725dd1c).  The compilation moved past the previous error but now fails at the link step.  The path to the netcdf library is not being defined (but it is able to find the netcdf include directory):
```
[100%] Linking Fortran executable chgres_cube.exe
ld: cannot find -lnetcdff
ld: cannot find -lnetcdf
[...]

PR #101, opened 2020-05-06
Update repository to use CMake

GeorgeGayno-NOAA:
Remove old build system and replace with a CMake-based system.

Remove any extensions from the executable files.


GeorgeGayno-NOAA:
Work was documented under issue #91.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
+cmake_minimum_required(VERSION 3.15)
+
+file(STRINGS "VERSION" pVersion)
+
+project(
+  ufs_util
+  VERSION ${pVersion}
+  LANGUAGES C Fortran)
+# LANGUAGES C CXX Fortran)
```
aerorahul, line 9: please remove commented code.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+if(NOT CMAKE_C_COMPILER_ID MATCHES "^(Intel|GNU)$")
+  message(WARNING "Compiler not officially supported: ${CMAKE_C_COMPILER_ID}")
+endif()
+
+find_package(PNG REQUIRED)
+find_package(ZLIB REQUIRED)
+find_package(Jasper REQUIRED)
+
+find_package(NetCDF REQUIRED C Fortran)
+#unset(CMAKE_C_COMPILER_LOADED)
```
aerorahul, line 34: remove commented code.

CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+#unset(CMAKE_CXX_COMPILER_LOADED)
+find_package(MPI REQUIRED )
+find_package(ESMF MODULE REQUIRED)
+find_package(WGRIB2 REQUIRED)
+
+find_package(OpenMP REQUIRED COMPONENTS Fortran)
```
aerorahul, line 40: Shouldn't threading be optional?

modulefiles/build.jet
```diff
@@ -1,26 +1,29 @@
...
- 
-export NETCDF_INCLUDE="-I${NETCDF}/include"
-export NETCDF_LDFLAGS_F="-L${NETCDF}/lib -lnetcdf -lnetcdff -L${HDF5}/lib -lhdf5 -lhdf5_fortran"
+module load g2/v3.1.0
 
-set FCMP ifort
+# Use DTCs version of esmf v8.  POC Dom H.
+module use -a /lfs3/projects/hfv3gfs/GMTB/modulefiles/intel-18.0.5.274
+module load esmf/8.0.0
+export NETCDF="/apps/netcdf/4.6.1/intel/18.0.5.274"
```
aerorahul, line 27: move this line to where NETCDF is loaded.

sorc/chgres_cube.fd/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+  set(CMAKE_Fortran_FLAGS_DEBUG
+      "-O0 -check -check noarg_temp_created -check nopointer -warn -warn noerrors -fp-stack-check -fstack-protector-all -fpe0 -debug -ftrapuv"
+  )
+elseif(CMAKE_Fortran_COMPILER_ID MATCHES "^(GNU|Clang|AppleClang)$")
+  set(CMAKE_Fortran_FLAGS "-g -fbacktrace -ffree-line-length-0 -fdefault-real-8 -fconvert=big-endian")
+  set(CMAKE_Fortran_FLAGS_RELEASE "-O3")
+  set(CMAKE_Fortran_FLAGS_DEBUG
+      "-O0 -ggdb -fno-unsafe-math-optimizations -frounding-math -fsignaling-nans -ffpe-trap=invalid,zero,overflow -fbounds-check"
+  )
+endif()
```
aerorahul, line 26: We need to move the default flags one level up for use with every application.
We should only add the application specific ones here to supplement the default flags.

sorc/fre-nctools.fd/shared_lib/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+  set(CMAKE_C_FLAGS "-traceback")
+  set(CMAKE_C_FLAGS_RELEASE "-O2")
+  set(CMAKE_C_FLAGS_DEBUG "-O0")
+elseif(CMAKE_C_COMPILER_ID MATCHES "^(GNU|Clang|AppleClang)$")
+  set(CMAKE_C_FLAGS " ")
+  set(CMAKE_C_FLAGS_RELEASE " ")
+  set(CMAKE_C_FLAGS_DEBUG " ")
+endif()
+
+add_library(shared_all STATIC ${c_src})
```
aerorahul, line 23: This directory is called `shared_lib`. To be consistent and easily relatable, call the library with the same name.

sorc/fre-nctools.fd/tools/fregrid/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+  set(CMAKE_C_FLAGS " ")
+  set(CMAKE_C_FLAGS_RELEASE " ")
+  set(CMAKE_C_FLAGS_DEBUG " ")
+endif()
+
+add_executable(fregrid ${c_src})
+
+target_link_libraries(fregrid
+  m
+  shared_all
```
aerorahul, line 22: `shared_lib` if you accept the previous suggestion.

sorc/fre-nctools.fd/tools/make_hgrid/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+if(CMAKE_C_COMPILER_ID MATCHES "^(Intel)$")
+  set(CMAKE_C_FLAGS "-traceback")
+  set(CMAKE_C_FLAGS_RELEASE "-O2")
+  set(CMAKE_C_FLAGS_DEBUG "-O0")
+    
+elseif(CMAKE_C_COMPILER_ID MATCHES "^(GNU|Clang|AppleClang)$")
+  set(CMAKE_C_FLAGS " ")
+  set(CMAKE_C_FLAGS_RELEASE " ")
+  set(CMAKE_C_FLAGS_DEBUG " ")
+endif()
```
aerorahul, line 17: This block is really repetitious and not necessary if the defaults are defined above.

sorc/global_chgres.fd/chgres_utils.f90
```diff
@@ -536,8 +536,8 @@ SUBROUTINE VINTG(IM,IX,KM1,KM2,NT,P1,U1,V1,T1,Q1,W1,P2,           &
 ! - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
 !  COMPUTE LOG PRESSURE INTERPOLATING COORDINATE                        
 !  AND COPY INPUT WIND, TEMPERATURE, HUMIDITY AND OTHER TRACERS         
-!$OMP PARALLEL DO DEFAULT(SHARED)                                       
-!$OMP+ PRIVATE(K,I)                                                     
+!$OMP PARALLEL DO DEFAULT(SHARED), &
```
aerorahul, line 539: Code changes included in the PR intended for build changes.
Please issue a separate PR for code related changes.

sorc/machine-setup.sh
```diff
@@ -28,6 +28,7 @@ if [[ -d /lfs3 ]] ; then
     target=jet
     module purge
     module use /lfs3/projects/hfv3gfs/nwprod/NCEPLIBS/modulefiles
+    module load cmake/3.16.1
```
aerorahul, line 31: please don't load modules in `machine-setup.sh`
There is a modulefile for each machine (or there was). This belongs there.
This file is to setup a clean module environment.

sorc/emcsfc_ice_blend.fd/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+  set(CMAKE_Fortran_FLAGS_RELEASE "-O3")
+  set(CMAKE_Fortran_FLAGS_DEBUG
+      "-O0 -ggdb -fno-unsafe-math-optimizations -frounding-math -fsignaling-nans -ffpe-trap=invalid,zero,overflow -fbounds-check"
+  )
+endif()
+
+set(exe_name emcsfc_ice_blend)
+add_executable(${exe_name} ${fortran_src})
+target_include_directories(
+  ${exe_name} PRIVATE)
```
DusanJovic-NOAA, line 21: This ``target_include_directories`` doesn't add any include directory to ``${exe_name}`` target. Consider removing it.

sorc/emcsfc_snow2mdl.fd/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+  set(CMAKE_Fortran_FLAGS_RELEASE "-O3")
+  set(CMAKE_Fortran_FLAGS_DEBUG
+      "-O0 -ggdb -fno-unsafe-math-optimizations -frounding-math -fsignaling-nans -ffpe-trap=invalid,zero,overflow -fbounds-check"
+  )
+endif()
+
+set(exe_name emcsfc_snow2mdl)
+add_executable(${exe_name} ${fortran_src})
+target_include_directories(
+  ${exe_name} PRIVATE)
```
DusanJovic-NOAA, line 26: Same as above.

sorc/chgres_cube.fd/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+    model_grid.F90
+    program_setup.f90
+    search_util.f90
+    static_data.F90
+    surface.F90
+    utils.f90
+    write_data.F90)
+
+if(CMAKE_Fortran_COMPILER_ID MATCHES "^(Intel)$")
+  set(CMAKE_Fortran_FLAGS "-g -traceback -r8 -i4 -convert big_endian -assume byterecl")
```
DusanJovic-NOAA, line 15: I think ``-i4`` flag is redundant. By default integers are 4-byte wide.

sorc/emcsfc_ice_blend.fd/CMakeLists.txt
GeorgeGayno-NOAA: I was able to remove this line from every CMakeLists.txt file except the one for chgres_cube (which needed the include directory for wgrib2).

sorc/emcsfc_snow2mdl.fd/CMakeLists.txt
GeorgeGayno-NOAA: Done.

sorc/chgres_cube.fd/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ endif()
 set(exe_name chgres_cube)
 add_executable(${exe_name} ${fortran_src})
 target_include_directories(
-  ${exe_name} PRIVATE ${ESMF_INC} ${WGRIB2_INCLUDES})
+  ${exe_name} PRIVATE ${WGRIB2_INCLUDES})
```
DusanJovic-NOAA, line 31: We should be able to express chgres_cube's dependency on WGRIB2 using imported target as we do with ESMF and all other libraries.

GeorgeGayno-NOAA: If someone wants to update FindWGRIB2.cmake, I will be happy to use it.

sorc/chgres_cube.fd/CMakeLists.txt
GeorgeGayno-NOAA: I removed all -i4 flags.

CMakeLists.txt
GeorgeGayno-NOAA: Done.

CMakeLists.txt
GeorgeGayno-NOAA: Done.

sorc/chgres_cube.fd/CMakeLists.txt
DusanJovic-NOAA: Try to add these 6 lines at the end of FindWGRIB2.cmake :
```
diff --git a/Modules/FindWGRIB2.cmake b/Modules/FindWGRIB2.cmake
index 702c0a4..5f3376b 100644
--- a/Modules/FindWGRIB2.cmake
+++ b/Modules/FindWGRIB2.cmake
@@ -18,3 +18,9 @@ else()
 endif()
 
 mark_as_advanced (WGRIB2_INCLUDES WGRIB2_LIBRARIES)
+
+add_library(wgrib2 UNKNOWN IMPORTED)
+set_target_properties(wgrib2 PROPERTIES
+  IMPORTED_LOCATION "${WGRIB2_LIBRARIES}"
+  INTERFACE_INCLUDE_DIRECTORIES "${WGRIB2_INCLUDES}"
+  INTERFACE_LINK_LIBRARIES "${WGRIB2_LIBRARIES}")
```
and then in ``sorc/chgres_cube.fd/CMakeLists.txt`` remove ``target_include_directories`` and replace ``${WGRIB2_LIBRARIES}`` with ``wgrib2``.

For example:
```
diff --git a/sorc/chgres_cube.fd/CMakeLists.txt b/sorc/chgres_cube.fd/CMakeLists.txt
index 0b888ce4..36509dcb 100644
--- a/sorc/chgres_cube.fd/CMakeLists.txt
+++ b/sorc/chgres_cube.fd/CMakeLists.txt
@@ -27,8 +27,8 @@ endif()
 
 set(exe_name chgres_cube)
 add_executable(${exe_name} ${fortran_src})
-target_include_directories(
-  ${exe_name} PRIVATE ${WGRIB2_INCLUDES})
+#target_include_directories(
+#  ${exe_name} PRIVATE ${WGRIB2_INCLUDES})
 target_link_libraries(
   ${exe_name}
   nemsio
@@ -38,8 +38,8 @@ target_link_libraries(
   sp_d
   w3nco_d
   esmf
+  wgrib2
   MPI::MPI_Fortran
-  ${WGRIB2_LIBRARIES}
   NetCDF::NetCDF_Fortran
   OpenMP::OpenMP_Fortran)
```

CMakeLists.txt
GeorgeGayno-NOAA: If making threads optional is a best practice/industry standard, I will make this change.

aerorahul: You can make it ON default, but there should be an option to not compile with threads.

kgerheiser: I encountered that here https://github.com/GeorgeGayno-NOAA/UFS_UTILS/issues/8

The call to `omp_get_max_threads()` needs to be conditional

GeorgeGayno-NOAA: I took a closer look at the global_chgres code.  Making threads optional in that code will require more work than it is worth.   Since global_chgres is nearly obsolete (replaced by chgres_cube), could the build process be modified to compile global_chgres only if the user turns on threads?  No one outside of EMC will ever use global_chgres.

kgerheiser: I think that sounds fine. Just add an if around the add_subdirectory to global_chgres? @aerorahul?

aerorahul:
```
if (OpenMP_FOUND)
  add_subdirectory(global_chgres.fd)
endif()
```

GeorgeGayno-NOAA: Great.  I will add the logic from @aerorahul.

sorc/fre-nctools.fd/shared_lib/CMakeLists.txt
GeorgeGayno-NOAA: Good idea.  Done.

sorc/fre-nctools.fd/tools/fregrid/CMakeLists.txt
GeorgeGayno-NOAA: Done.

sorc/machine-setup.sh
GeorgeGayno-NOAA: Good point.  Done.

modulefiles/build.jet
GeorgeGayno-NOAA: Done.

CMakeLists.txt
GeorgeGayno-NOAA: The threaded routines in global_chgres are also used by global_cycle.  So updating global_cycle to be thread optional will be more work than it is worth right now.  So logic was added to not build global_chgres **and** global_cycle if the user turns off threads.  No one outside EMC is using global_cycle currently and it will not be part of the next public release.

sorc/chgres_cube.fd/CMakeLists.txt
GeorgeGayno-NOAA: Do you mean move CMAKE_Fortran_FLAGS and CMAKE_C_FLAGS to the root level CMakeLists.txt file?

sorc/global_chgres.fd/chgres_utils.f90
GeorgeGayno-NOAA: > Code changes included in the PR intended for build changes.
> Please issue a separate PR for code related changes.

I can check, but I only intended to change the syntax of the OpenMP directives to satisfy the latest version of the Intel compiler.

sorc/chgres_cube.fd/CMakeLists.txt
aerorahul: I meant all of them
```
CMAKE_Fortran_FLAGS
CMAKE_Fortran_FLAGS_RELEASE
CMAKE_Fortran_FLAGS_DEBUG
```

and the C counterparts.

After that, at this level, we can have
```
set( CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} --any-application-specifc-flags" )
```
Similarly for RELEASE and DEBUG (depending on the compiler set)
This eliminates repetition. If stock options are good, then the entire section is gone.

GeorgeGayno-NOAA: Done.

sorc/fre-nctools.fd/tools/make_hgrid/CMakeLists.txt
GeorgeGayno-NOAA: Done.

sorc/global_chgres.fd/chgres_utils.f90
GeorgeGayno-NOAA: Some background:  The old build system applied the '-qopenmp' flag to a subset of the global_chgres routines.  For some reason, chgres_utils.f90 was not compiled with that flag.  The CMake build compiles all routines with -qopenmp.  And the compiler complained about the old syntax.

aerorahul: Ok. I can understand this change and am willing to slide.

reg_tests/chgres_cube/driver.dell.sh
```diff
@@ -27,8 +27,8 @@ module load EnvVars/1.0.2
 module load ips/18.0.1.163
 module load impi/18.0.1
 module load lsf/10.1
-module use /usrx/local/dev/modulefiles
-module load NetCDF/4.5.0
+module use /usrx/local/nceplibs/dev/NCEPLIBS/modulefiles
+module load netcdf_parallel/4.7.4
```
aerorahul, line 31: why are modules being loaded in the driver.XXX.sh? Isn't sourcing the machine appropriate modulefile enough?
The modulefile should load all the modules required.

sorc/chgres_cube.fd/CMakeLists.txt
GeorgeGayno-NOAA: I tested your change and it works.  I can initiate a pull request for the updated FindWGRIB2.cmake.

reg_tests/chgres_cube/driver.dell.sh
GeorgeGayno-NOAA: That is a good point.  I will make that change.

sorc/chgres_cube.fd/CMakeLists.txt
GeorgeGayno-NOAA: Updated the branch to point to Dusan's improved FindWGRIB2.cmake.

sorc/global_chgres.fd/chgres_utils.f90
GeorgeGayno-NOAA: > Code changes included in the PR intended for build changes.
> Please issue a separate PR for code related changes.
> 
> I can check, but I only intended to change the syntax of the OpenMP directives to satisfy the latest version of the Intel compiler.

Some follow up:  I ran a canned `global_chgres` case on Dell using 14 threads.  The branch (28a11c9d) and develop (270f9dc) produced identical output (according to the nccmp utility).   So I am confident I did not introduce any errors into the code.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA CMake changes are good with me.
> I am ok with the code-changes too, but in the future, a separate PR would be appropriate.
> I also made a comment on loading modulefiles in the driver script.
> If it is not too  much work, please address, or leave for another time.
> Thanks for the clean-up of the build system.

All driver scripts were updated to load the build modules.

GeorgeGayno-NOAA:
Thanks everyone for the EXCELLENT comments and for your help getting CMake to work.  I will rerun all the regression tests, then do the merge.

driver_scripts/driver_grid.dell.sh
```diff
@@ -56,8 +56,9 @@ module load EnvVars/1.0.2
 module load lsf/10.1
 module load ips/18.0.1.163
 module load impi/18.0.1
-module load NetCDF/4.5.0
 module load HDF5-serial/1.10.1
+module use /usrx/local/nceplibs/dev/NCEPLIBS/modulefiles
+module load netcdf_parallel/4.7.4
 module list
```
aerorahul, line 62: may be this is an oversight, but this script is still loading modules explicitly. Shouldn't it be sourcing modulefiles?

reg_tests/chgres_cube/driver.dell.sh
```diff
@@ -22,14 +22,9 @@
 
 set -x
 
-module purge
-module load EnvVars/1.0.2
-module load ips/18.0.1.163
-module load impi/18.0.1
+source ../../sorc/machine-setup.sh > /dev/null 2>&1
+source ../../modulefiles/build.$target
 module load lsf/10.1
```
aerorahul, line 27: This module load is feeling lonely. Can you pull that into the modulefile?

reg_tests/global_cycle/driver.dell.sh
```diff
@@ -32,15 +32,9 @@
 
 set -x
 
-module purge
-module load EnvVars/1.0.2
-module load ips/18.0.1.163
+source ../../sorc/machine-setup.sh > /dev/null 2>&1
+source ../../modulefiles/build.$target
 module load lsf/10.1
```
aerorahul, line 37: same here.

reg_tests/grid_gen/driver.dell.sh
```diff
@@ -19,14 +19,9 @@
 #
 #-----------------------------------------------------------------------------
 
-module purge
-module load EnvVars/1.0.2
+source ../../sorc/machine-setup.sh > /dev/null 2>&1
+source ../../modulefiles/build.$target
 module load lsf/10.1
```
aerorahul, line 24: and here.

reg_tests/chgres_cube/driver.cray.sh
```diff
@@ -22,14 +22,7 @@
...
 
-module purge
-module load PrgEnv-intel/5.2.56
-module rm intel
-module load intel/16.3.210
-module load cray-mpich/7.2.0
-module load craype-haswell
-module load cray-netcdf/4.3.3.1
-module load xt-lsfhpc/9.1.3
+source ../../modulefiles/build.wcoss_cray
```
aerorahul, line 25: you need to source machine-setup.sh before this line.

driver_scripts/driver_grid.dell.sh
GeorgeGayno-NOAA: Good catch.

GeorgeGayno-NOAA: Fixed.

reg_tests/chgres_cube/driver.dell.sh
GeorgeGayno-NOAA: Fixed.

reg_tests/global_cycle/driver.dell.sh
GeorgeGayno-NOAA: Fixed.

reg_tests/grid_gen/driver.dell.sh
GeorgeGayno-NOAA: Fixed.

modulefiles/build.wcoss_dell_p3
```diff
@@ -1,25 +1,31 @@
...
 module load sfcio/1.0.0
+module load sigio/2.1.0
 module load gfsio/1.1.0
 module load landsfcutil/2.1.0
-module load bacio/2.0.2
-module load  sigio/2.1.0 
+module load nemsiogfs/2.0.1
+module load g2/3.1.0
+module load jasper/1.900.29
+export Jasper_ROOT="/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29"
```
aerorahul, line 23: what does `module load jasper/1.900.29` do if you have to manually export `Jasper_ROOT`?
If the module loads an environment variable e.g. `JASPER`, it is best to use that to set it to `Jasper_ROOT`, rather than specifying the full path.

aerorahul: This is going to be harder than simply setting the variable because of it being in a module file. We should ask the sysadmin to define this variable in the jasper modulefile itself.
I take this comment back.

aerorahul:
@GeorgeGayno-NOAA 
Thank you for all the hard-work and persevering through the comments and reviews.  `UFS_UTILS` is in good shape and ready for merge when you are satisfied with the testing on various platforms.

modulefiles/build.wcoss_dell_p3
GeorgeGayno-NOAA: Here is the output from the Dell module.  Many of our modules don't provide what CMake wants.  And I don't always know the best way to handle it.  Should I not load the module at all and just hardwire what I need?  Should I try to create Jasper_ROOT from one of the exiting variables?

```prepend_path("PATH","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29/bin")
prepend_path("LD_LIBRARY_PATH","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29/lib")
prepend_path("MANPATH","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29//man")
setenv("JASPER_INC","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29/include")
setenv("JASPER_LIBDIR","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29/lib")
setenv("JASPER_LIB","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29/lib/libjasper.a")
setenv("JASPER_SRC","/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29/src/jasper-1.900.29")
setenv("JASPER_VER","1.900.29")

Comment (aerorahul) [reply modulefiles/build.wcoss_dell_p3:23]: @GeorgeGayno-NOAA Leave as is for now and open a ticket with WCOSS helpdesk and ask them to add a line in their modulefile for Jasper as follows to allow CMake to find the correct Jasper installation:
```
setenv("Jasper_ROOT", "/usrx/local/prod/packages/gnu/4.8.5/jasper/1.900.29")
```
When they agree and make the change, we can remove the `export Jasper_ROOT= ...` line from our modulefile.

DusanJovic-NOAA:
All utilities compiled successfully on my laptop, and I also tested the execution of ``chgres_cube`` for C96 case. Ready for merge.

Issue #22, opened 2019-10-04
Include 2D coordinate arrays (geolat and geolon) and associated variable attributes in orog.fd, chgres_cube.fd, and sfc_climo_gen.fd for correct output viewing in ncview


JeffBeck-NOAA:

JeffBeck-NOAA:
Gerard and I would like to work on this during the next couple weeks

GeorgeGayno-NOAA:
I think everything is working.  Using cd0aafc on Hera, I created a C96 uniform global grid.  The 'oro' and 'surface' files had the required geo-referencing records.  I was able to add a map background when viewing them with 'ncview'.  I fed the 'oro' and 'surface' file into chgres_cube for a Feb 2, 2020 test case.  It ran without issue.  And the files output from chgres_cube had the required geo-referencing records, which were recognized by 'ncview'.  All files were then fed to the forecast model, which ran successfully for a 24-hour forecast.

GeorgeGayno-NOAA:
After the reviewers gave their blessing, the branch at a3ea604 was compiled on Hera, Jet, Dell and Cray.  Then all regression tests were run.  As expected the `chgres_cube` and `grid_gen` regression tests failed.  The output files did not match the official baseline according to the `nccmp` utility.  However, the differences were due solely to the additional metadata and lat/lon records in the files.   All other records were identical, as they should be.

Will merge to the authoritative repository.

GeorgeGayno-NOAA:
Merged to 'develop' at 7448e0a.  Closing ticket.

Issue #11, opened 2019-09-06
Missing values in the coordinate lon lat variables in the regional orog file


BinLiu-NOAA:
There are some strange missing values (see the 0.9969E+37 values below) in the two netcdf coordinate variables (lon and lat) in the orography files for the regional FV3SAR and HAFS configurations, e.g., C768_oro_data.tile7.halo0.nc. The geolon and geolat variables, however, look normal.

Not sure whether these missing coordinate lon/lat values matter or not. But it might be better to fill them with reasonable values or just use npx, npy as coordinate variables.

BTW, when checking the lon, lat values in the sfc_data.tile7.nc and gfs_data.tile7.nc files (generated by chgres), they all look normal though.

Variable: f
Type: file
filename:   oro_data
path:   oro_data.nc
   file global attributes:
   dimensions:
      lon = 2560
      lat = 2160
   variables:
      float lon ( lon )
         cartesian_axis :   X

      float lat ( lat )
         cartesian_axis :   Y

      float geolon ( lat, lon )

      float geolat ( lat, lon )

      float slmsk ( lat, lon )

      float land_frac ( lat, lon )

      float orog_raw ( lat, lon )

      float orog_filt ( lat, lon )

      float stddev ( lat, lon )

      float convexity ( lat, lon )

      float oa1 ( lat, lon )

      float oa2 ( lat, lon )

      float oa3 ( lat, lon )

      float oa4 ( lat, lon )

      float ol1 ( lat, lon )

      float ol2 ( lat, lon )

      float ol3 ( lat, lon )

      float ol4 ( lat, lon )

      float theta ( lat, lon )

      float gamma ( lat, lon )

      float sigma ( lat, lon )

      float elvmax ( lat, lon )

Coordinates:

     lon:

   337.1       337.1       337.0       337.0       337.0       336.9
   336.9       336.9       336.9       336.8       336.8       336.8
   336.8       336.7       336.7       336.7       336.7       336.6
   336.6       336.6       336.6       336.5       336.5       336.5
   336.4       336.4       336.4       336.4       336.3       336.3
   336.3       336.3       336.2       336.2       336.2       336.2
   336.1       336.1       336.1       336.0       336.0       336.0
...
   246.3       246.3       246.3       246.2       246.2       246.2
  0.9969E+37  0.9969E+37  0.9969E+37  0.9969E+37  0.9969E+37  0.9969E+37
[...]

PR #109, opened 2020-06-03
Add geo-referencing to 'oro', 'sfc_climo' and 'coldstart' files

GeorgeGayno-NOAA:
Add geo-referencing to the files output by the orography maker, the sfc_climo_gen program, and chgres_cube.  Remove lat/lon records from the orography files, which are not used and can contain bad values.  See issues #11 and #22.

### DIFF
--- a/sorc/chgres_cube.fd/write_data.F90
+++ b/sorc/chgres_cube.fd/write_data.F90
@@ -1643,6 +1643,8 @@
 
  use model_grid, only            : num_tiles_target_grid, &
                                    landmask_target_grid, &
+                                   latitude_target_grid, &
+                                   longitude_target_grid, &
                                    i_target, j_target, lsoil_target
 
  use program_setup, only         : convert_nst, halo=>halo_bndy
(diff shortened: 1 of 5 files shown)


GeorgeGayno-NOAA:
> Code changes look good.  The only thing I'll note is that when I was ncview-ing a C192 regional grid orog file that @GeorgeGayno-NOAA pointed me to, the map was upside down (north-south flipped).  Didn't get around to looking to see if there is an easy fix in ncview (I don't think there is).  If there's an easy fix you can make in the fortran code, that would be great.

The (non-Jim Purser) regional CONUS grids are always upside-down.  That is how GFDL oriented the six global tiles.   When I view in ncview, the map background is also upside-down, so it is consistent.  Are you seeing a flip between the map background and the grid data?

gsketefian:
@GeorgeGayno-NOAA The map background is consistent with the fields, so that part is ok.  We can fix this in another PR if users are bothered by it.  We can wait and see.

GeorgeGayno-NOAA:
Thanks for your reviews.  I will run the regression test suite on our supported machines, then do the merge.

Issue #111, opened 2020-06-08
Add lake maker to orography data


GeorgeGayno-NOAA:
Adding lake maker from Shan Sun and Ning Wang: generate lake fraction and depth on the FV3 grid, add them to oro_data, and adjust land_frac and slmsk, such that land_frac and slmsk are consistent with lake_frac.

GeorgeGayno-NOAA:
Can you define 'slmsk', 'land_frac' and 'lake_frac' and how they relate to each other?  Should 'land_frac' and 'lake_frac' add to 1.0?

GeorgeGayno-NOAA:
The branch at 85c0602 was successfully compiled on WCOSS-Cray/Dell, Hera, Jet and Orion.   A C96 global uniform grid with lakes was successfully created on these machines as well.  And the grid_gen regression test passed on all machines.  Will merge.

GeorgeGayno-NOAA:
Merged to 'develop' at 7770042.  Deleting branch and closing issue.

PR #112, opened 2020-06-09
Adding lake maker

shansun6:
Generating lake fraction and depth on the FV3 grid and adding them to oro_data. This commit references issue #111.


GeorgeGayno-NOAA:
@shansun6  The addition of lakes should be a part of the grid generation process, not a stand-alone process.  The main grid driver script is ./fv3gfs_driver_grid.sh.  It is called by machine-specific batch scripts in ./driver_scripts.

Issue #115, opened 2020-06-15
NDATE not defined in gdas init utility scripts


GeorgeGayno-NOAA:
During a recent update the 'develop' (3ad7d83), the load of 'prod_util' was inadvertently removed.  As a result, NDATE is undefined in the GDAS initialization utility scripts.  NDATE is used when processing GFS v15 data.

GeorgeGayno-NOAA:
Tested the branch at 32779b3 on Hera, Cray and Dell as follows:

- The entire repository compiled without error.
- All regression tests passed.
- Successfully ran the GDAS initialization script for a July 19, 2019 test case (GFS v15).  NDATE was properly defined and all coldstart files were created.

Will merge to 'develop'.

GeorgeGayno-NOAA:
Merged to 'develop' at 10426d4.  Closing ticket.

PR #117, opened 2020-06-16
Define NDATE in GDAS initialization scripts

GeorgeGayno-NOAA:
The NDATE program was not being defined in the GDAS initialization scripts.
As a result, the scripts would not work when initializing with GFS v15 data.

For details see #115.

### DIFF
--- a/modulefiles/build.wcoss_dell_p3
+++ b/modulefiles/build.wcoss_dell_p3
@@ -29,3 +29,6 @@ module use /gpfs/dell2/emc/modeling/noscrub/emc.nemspara/soft/modulefiles
 module load esmf/8.0.0_ParallelNetCDF
 
 export WGRIB2_ROOT=/gpfs/dell2/emc/modeling/noscrub/George.Gayno/wgrib2
+
+module use /usrx/local/dev/modulefiles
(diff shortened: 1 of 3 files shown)


modulefiles/build.wcoss_dell_p3
```diff
@@ -29,3 +29,6 @@ module use /gpfs/dell2/emc/modeling/noscrub/emc.nemspara/soft/modulefiles
 module load esmf/8.0.0_ParallelNetCDF
 
 export WGRIB2_ROOT=/gpfs/dell2/emc/modeling/noscrub/George.Gayno/wgrib2
```
KateFriedman-NOAA, line 31: Could loading the grib_util modules (which contains $WGRIB2) replace this line or is this a special wgrib2 you're testing?

GeorgeGayno-NOAA: This is not the wgrib2 program.  It is the library API version of wgrib2 that is callable from Fortran.  We hope to formally implement it shortly (it is required for GFS v16).

KateFriedman-NOAA: Got it, thanks!

PR #128, opened 2020-08-10
Adding lake maker (replacing PR #121)

shansun6:
Generating lake fraction and depth on the FV3 grid, add them to oro_data, and adjust lake_frac and land_frac, such that the two arrays are consistent.

This refers to issue # 111.


GeorgeGayno-NOAA:
@shansun6  There is no need to open new pull requests.  You can update your branch from 'develop' under an existing PR.

GeorgeGayno-NOAA:
@shansun6 I had to make some changes to run on our Cray.  Will send you a pull request soon.

GeorgeGayno-NOAA:
@shansun6  sfc_climo_gen is the last program run when creating a grid.  It sets surface fields, such as vegetation type, at land points.  It uses 'slmsk' to determine what is land.  How should the program identify land when the lake maker is used?

shansun6:
Hi George,

Lake maker is based on GLDBv2 (Choulga et al. 2014) in the 1st step. In the
2nd step, lake_frac and land_frac (in the oro data) are reconciled in such
a way:
        if (lake_frac(i) >= lake_cutoff) then ! non-zero lake_frac
dominates over land_frac
          land_frac(i) = max(0., min(1., 1.-lake_frac(i)))
        elseif (inland(i) == 1.) then ! land_frac dominates over lake_frac
at inland points
          lake_frac(i) = max(0., min(1., 1.-land_frac(i)))
        end if

before lake_frac and lake_depth are added to oro data. land_frac and slmsk
in oro_data are modified as well (slmsk(i) = nint(land_frac(i)). All these are
done in /sorc/orog.fd/lake.fd/lakefrac.F90.

Please let me know if you have any questions.

Thanks,
Shan

GeorgeGayno-NOAA:
The program that creates these files: /scratch1/NCEPDEV/global/glopara/fix/fix_fv3_gmted2010/C96/fix_sfc, assumes a point is all land or all non-land.  It does not account for partial land.  So it does not need to be changed?

https://github.com/NOAA-EMC/UFS_UTILS/tree/develop/sorc/sfc_climo_gen.fd

shansun6:
Hi George,

Yes, you are correct: once oro data is made, the rest of the
programs downstream from here should read "land_frac", which has a more
accurate surface type, rather than "slmsk", which is nint(land_frac). I'd
suggest removing "slmsk" from oro_data, simply because slmsk can be derived
from land_frac.

Our focus has been around oro data where we have the expertise: generating
lake frac/depth and modifying land_frac/slmsk so they are consistent with
lake_frac in oro data. We'd suggest the modification of surface-related
routines to its experts, and be happy to answer all the questions related
to oro data. Thanks for your understanding.

Shan

GeorgeGayno-NOAA:
@shansun6  Since sfc_climo_gen is run as part of the grid generation scripts, I will update it for partial land/non-land points under this pull request.

If 'slmsk' is a meaningless record when using the lake maker, then go ahead and remove it.  (But we need to keep it when the lake maker is not used).

shansun6:
Hi George,

Thanks for working on sfc_climo_gen. Since the current PR is about lake
maker, is it possible to have sfc_climo_gen under a separate PR?

I believe "slmsk" is set to nint(land_frac) with or without lake maker. It
is a required variable for the nonfractional FV3 to read in as a real
variable (FV3, when running at in the fractional mode, reads in land_frac
instead). FV3 code can be changed to read land_frac instead of slmsk in the
nonfrac case. I will discuss it with Jun for that. Then we will drop slmsk
from oro_data. Will keep you posted,

Thanks,
Shan

GeorgeGayno-NOAA:
@shansun6 The changes required for sfc_climo_gen are very minor.  About 10 lines of code.  And I have it working.  So, it should not be a problem to include it in this PR.

GeorgeGayno-NOAA:
@shansun6   When testing sfc_climo_gen, I set land fields such as vegetation type at all points with a non-zero land fraction.  Some islands off South America (a C96 case) had a land fraction less than 1%.  Will the land and ocean physics be run at these points?  Is it worth running the land physics for a point with < 1% land?

shansun6:
Great that you are making progress already. Yes, land fields will be needed
at all points with a non-zero land fraction.

It has been in discussion what the minimum land frac should be. At the
moment, the minimum land frac in the to-be-committed code is set to be
1.e-6 (0.0001%). We have been using 1.e-4 in the fractional grid. A smaller
number is meant to capture a more authentic coastline and islands. What
number would you prefer? I will see what the land group says.

Thanks,
Shan

GeorgeGayno-NOAA:
@shansun6  I tested your changes on all officially supported machines using my branch of the same name.  My branch has all your updates and includes changes to run on all machines and minor clean up.  I will close this PR and open another one.

PR #142, opened 2020-09-02
add tripoletools folder with 3 submodules

DeniseWorthen:
1) CICE5_ICgen: grid generation for cice5/cice6

2) CICE5_gridgen: initial condition generation for cice5/cice6

3) WeightGen: weight generation

a) gen_fixgrid and associated fortran code:
a fortran routine to read the MOM6 supergrid
and create a netcdf file which is the basis for the
remaining transformations

b) generate_iceocn_weights.ncl:
esmf weight generation for tripole:tripole or
tripole:rectilinear or other grid:grid transformations

c) generate_icemesh.ncl:
generation of the ice_mesh for use by the CMEPS mediator

d) generate_frac_land_weights.sh:
generation of the frac_land weights (the ocean mask on the
fv3 grid) using ESMF_RegridWeightGen

d) make_frac_land.ncl:
creation of the 6-tiled land_frac files

Note: these need to be moved into the appropriate places and adjusted for general use. I am happy to work with you on the best way to do this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,12 @@
 	path = cmake
 	url = https://github.com/NOAA-EMC/CMakeModules
 	branch = develop
+[submodule "tripoletools/CICE5_gridgen"]
+	path = tripoletools/CICE5_gridgen
+	url = https://github.com/DeniseWorthen/CICE5_gridgen
+[submodule "tripoletools/CICE5_ICgen"]
+	path = tripoletools/CICE5_ICgen
+	url = https://github.com/DeniseWorthen/CICE5_ICgen.git
+[submodule "tripoletools/WeightGen"]
+	path = tripoletools/WeightGen
+	url = https://github.com/DeniseWorthen/WeightGen.git


DeniseWorthen:
FYI: @junwang-noaa @arunchawla-NOAA

GeorgeGayno-NOAA:
@DeniseWorthen So you are simply adding your code as submodules?  Why does UFS_UTILS need to be involved?

DeniseWorthen:
These need to be integrated into ufs-utils. I do not know where they best belong, or what structure you are using. They also need to be worked with the be of use to a general user. It was my impression that you were expecting to see this as a PR so that we could work together to best figure out how they should be incorporated into ufs-utils.

GeorgeGayno-NOAA:
> These need to be integrated into ufs-utils. I do not know where they best belong, or what structure you are using. They also need to be worked with to be of use to a general user. It was my impression that you were expecting to see this as a PR so that we could work together to best figure out how they should be incorporated into ufs-utils.

Denise, the PR is always the last step after everything has been tested.  Open an issue.  Then, we can discuss the details.  Keep your branch.  But close this PR.

PR #146, opened 2020-09-10
Add chgres_cube support for regional model (HRRR, RAP, and NAM) grib2 files and UFS_UTILS build support for Cheyenne, Stampede2, and Odin

JeffBeck-NOAA:
This PR represents required UFS_UTILS changes for the SRW App release:

- Adds support in chgres_cube for regional grib2 files for HRRR, RAP, and NAM models
- Adds support in chgres_cube to interpolate from nine to four soil layers (initialize from nine-layer RUC-LSM RAP/HRRR data for CCPP SDFs using four-layer NOAH/NOAH MP)
- Adds UFS_UTILS build/module support for Cheyenne, Stampede2, and Odin HPC systems, which are Tier 1 and Tier 2 platforms for the SRW App release
- Modifies all module files to use "setenv" instead of "export" commands and modifies build_all.sh script to use "module use/load" instead of "source" to allow module files to be loaded at both build and run time
- Adds logic to chgres_cube write_data.F90 file to output tile7 NetCDF files when "regional" flag is on
- Adds support in chgres_cube to allow for either a dot or an underscore when reading in the mosaic file (C768.mosaic.halo4.nc or C768_mosaic.halo4.nc)

Both build and run time changes have been tested successfully on Hera and Jet (so far) using the SRW App community workflow.   The build has also been tested on Cheyenne, Stampede2, and Odin, with full end-to-end workflow tests to commence shortly.  Testing on WCOSS is pending.

The following developers have provided code for this PR: @LarissaReames-NOAA, @JeffBeck-NOAA, @ywangwof, @gsketefian, @JulieSchramm, and @mkavulich


climbfuji:
@JeffBeck-NOAA your description says that you modified the modulefiles to become "lua" modules and that you are loading them with `module load` instead of sourcing them. This will not work on gaea, which doesn't know lua - need tcl/tk modules.

JeffBeck-NOAA:
@DomHeinzeller, how are module files loaded on Gaea?  Nothing has changed except using "module load" instead of "source" during the build and "setenv" instead of "export" in those module files.

climbfuji:
> @DomHeinzeller, how are module files loaded on Gaea? Nothing has changed except using "module load" instead of "source" during the build and "setenv" instead of "export" in those module files. So, they're not technically "lua" module files, the "setenv" command is just "lua-approved".

So you are saying that these modulefiles are tcl/tk? Then it's ok, gaea can handle those via `module load`.

GeorgeGayno-NOAA (changes requested):
Jeff,  I see two large binary files under ./fix_chgres.  We can only baseline source code.  Those files will need to be stored locally on each machine.

JeffBeck-NOAA:
> Jeff, I see two large binary files under ./fix_chgres. We can only baseline source code. Those files will need to be stored locally on each machine.

@GeorgeGayno-NOAA Both fixed files have now been removed from the PR (one commit I made, and another I merged in from Larissa's fork).  They will be placed in the fixed file directories on Tier 1 platforms and on the EMC FTP server for users who don't have access to NOAA HPC.

sorc/chgres_cube.fd/run/config.C768.odin.hrrr.nml
```diff
@@ -0,0 +1,42 @@
+# Takes lots of memory.  Use two nodes - six tasks per node.
```
GeorgeGayno-NOAA, line 1: @JeffBeck-NOAA I removed this run directory a while back as I did not want scripts and config files in the sorc directory.

modulefiles/build.orion
```diff
@@ -21,7 +21,7 @@ module load nemsiogfs/2.5.3
 module load landsfcutil/2.4.1
```
GeorgeGayno-NOAA, line 21: @JeffBeck-NOAA  Had to change the first line to #%Module### to get it to compile on Orion.

GeorgeGayno-NOAA:
Compiled c51b5ae on Hera, then ran the chgres_cube regression tests.  The surface files were different for all tests.  Some of the surface fields, such as veg type, had numerous missing values.

sorc/chgres_cube.fd/run/config.C768.odin.hrrr.nml
JeffBeck-NOAA: @GeorgeGayno-NOAA, I pushed a commit to remove this directory.

modulefiles/build.orion
JeffBeck-NOAA: @GeorgeGayno-NOAA, I pushed a commit to update the build.orion file.

JeffBeck-NOAA:
> Compiled [c51b5ae](https://github.com/NOAA-EMC/UFS_UTILS/commit/c51b5aea283aa507e14b170ad355b6e812f05708) on Hera, then ran the chgres_cube regression tests. The surface files were different for all tests. Some of the surface fields, such as veg type, had numerous missing values. See /scratch2/NCEPDEV/stmp1/George.Gayno/chgres_reg_tests.bad.sfc

@GeorgeGayno-NOAA, @LarissaReames-NOAA and I are working on the surface problems.

GeorgeGayno-NOAA:
I see several new routines in surface.F90, such as "replace_land_sfcparams", "check_smois_land", "check_smois_water".  What are these routines correcting?

sorc/chgres_cube.fd/surface.F90
```diff
@@ -2234,6 +2612,498 @@ subroutine calc_liq_soil_moisture
...
+ do i =clb(1),cub(1)
+   do j = clb(2),cub(2)
+     if (landmask_ptr(i,j)==1 .and. soilm_target_ptr(i,j,1) > 0.75) then
+       !write(*,'(a,2i5,a,i3,f5.2,2i3)') "CORRECTING G.P.",i,j," FROM LAND TO SEA VALUES; &
+       !    curr landmask, soilm, stype, vtype = ",landmask_ptr(i,j),&
+       !    soilm_target_ptr(i,j,1),nint(soil_type_target_ptr(i,j)), &
+       !    nint(veg_type_target_ptr(i,j))
+       soil_type_target_ptr(i,j) = 0.0
+       veg_type_target_ptr(i,J) = 0.0
+       landmask_ptr(i,j) = 0
```
GeorgeGayno-NOAA, line 2731: I would not change the land mask in chgres_cube.  The coldstart files will not match the 'orog' files.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -2234,6 +2612,498 @@ subroutine calc_liq_soil_moisture
...
+
+ print*,"- CALL FieldGet FOR TARGET GRID SNOW ALBEDO."
+ call ESMF_FieldGet(mxsno_albedo_target_grid, &
+                    farrayPtr=mxsno_albedo_target_ptr, rc=rc)
+ if(ESMF_logFoundError(rcToCheck=rc,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
+    call error_handler("IN FieldGet", rc)
+
+ do i =clb(1),cub(1)
+   do j = clb(2),cub(2)
+     if (landmask_ptr(i,j)==1 .and. soilm_target_ptr(i,j,1) > 0.75) then
```
GeorgeGayno-NOAA: Why would land points have bad soil moisture values after interpolation to the target grid?  Is the input grib2 landmask wrong?

sorc/chgres_cube.fd/surface.F90
JeffBeck-NOAA: @GeorgeGayno-NOAA This is old code that, due to other routines, never gets called anymore.  We will comment it out and issue a commit to this PR.

LarissaReames-NOAA: If you'll look in the driver subroutine, the call to that subroutine is commented out. I'm going to comment out the entire subroutine for now. I'm reticent to remove it completely just in case we need it again.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -2234,6 +2612,498 @@ subroutine calc_liq_soil_moisture
...
+
+ print*,"- CALL FieldGet FOR TARGET GRID SNOW ALBEDO."
+ call ESMF_FieldGet(mxsno_albedo_target_grid, &
+                    farrayPtr=mxsno_albedo_target_ptr, rc=rc)
+ if(ESMF_logFoundError(rcToCheck=rc,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
+    call error_handler("IN FieldGet", rc)
+
+ do i =clb(1),cub(1)
+   do j = clb(2),cub(2)
+     if (landmask_ptr(i,j)==1 .and. soilm_target_ptr(i,j,1) < 0.001) then !.and. &
```
GeorgeGayno-NOAA: What is the cause of these bad points?

LarissaReames-NOAA: At GSL's request, we've added in functionality that allows the user to read in vegetation and soil type from the input grid. They wanted the higher resolution data available from HRRR files. This resulted in some mismatched land/water points. Note that all of these subroutines are now only called if the  *_from_climo flags are set to "False", and thus the program reads in these new variables. These new sub routines were required to deal with these problems. You'll find other code in the surface interp subroutine related to these flags as well.

JeffBeck-NOAA:
> I see several new routines in surface.F90, such as "replace_land_sfcparams", "check_smois_land", "check_smois_water". What are these routines correcting?

@GeorgeGayno-NOAA, as @LarissaReames-NOAA mentioned elsewhere, reading the additional grib2 external model data will sometimes result in mismatches for land surface variables that require target grid modifications to certain fields to ensure that they match the target landmask.

sorc/chgres_cube.fd/surface.F90
JeffBeck-NOAA: This was a mismatch problem between the source and target landmasks, resulting in soil moisture values being inconsistent with the target landmask value (land, water, or ice).

JeffBeck-NOAA:
@GeorgeGayno-NOAA, @LarissaReames-NOAA and I have finalized our changes to this PR to resolve the regression test failures.  We reran the tests and confirmed that there are no longer any discrepancies between the baseline and resulting surface file variables.  Note that there are metadata and/or file name differences that still cause the regression tests to fail (for example, tile7 from the new chgres_cube executable is named tile1 for the baseline directory data files).  Please feel free to rerun the regression tests.

sorc/chgres_cube.fd/surface.F90
GeorgeGayno-NOAA: Why would you need it?  If your input file has a proper mask and good data, then you should be able to map to any target/model grid.  You should never have to change the target land mask.

sorc/chgres_cube.fd/surface.F90
GeorgeGayno-NOAA: This adjustment is not required for other input data sources.  What exactly is mismatched?  Is the input data mask and soil moisture mismatched?  The maximum value of soil moisture is around 0.5.

sorc/chgres_cube.fd/surface.F90
LarissaReames-NOAA: It was a quick fix that was implemented to a problem we were having with soil moisture mismatches just to see if we could fix the problem. We recognized that this sort of solution was not appropriate and implemented other more appropriate fixes elsewhere in the code. Again, this code is no longer used.

sorc/chgres_cube.fd/surface.F90
LarissaReames-NOAA: See my comment below about changes made at GSL's request.

sorc/chgres_cube.fd/surface.F90
GeorgeGayno-NOAA: Are these fields read in from the 'geogrid' file?  Can I look at this file?  Is it on Hera?

LarissaReames-NOAA: They are in the HRRRX files that GSL uses for their HRRR-like runs. @JeffBeck-NOAA should be able to point you to the  correct file.

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -5604,6 +5950,180 @@ SUBROUTINE READ_FV3_GRID_DATA_NETCDF(FIELD,TILE_NUM,IMO,JMO,LMO, &
...
+        call error_handler("IN FieldGather", error)
+   print*,"- CALL FieldGather FOR INPUT GRID LATITUDE"
+   call ESMF_FieldGather(latitude_input_grid, lat, rootPet=0, tile=1, rc=error)
+   if(ESMF_logFoundError(rcToCheck=error,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
+        call error_handler("IN FieldGather", error)
+
+   if (localpet==0) then
+     print*,"- CALCULATE ROTATION ANGLE FOR ROTATED_LATLON INPUT GRID"
+     cmdline_msg = "wgrib2 "//trim(file)//" -d 1 -grid &> temp.out"
+     call system(cmdline_msg)
```
GeorgeGayno-NOAA, line 6017: These system calls are kind of clunky.  Do they work on all platforms?  There may be a cleaner way to return the map information.

LarissaReames-NOAA: They should work on all platforms, yes. The path to the wgrib2 executable is included in the path by default in the release configuration for all primary supported platforms. I'm open to other suggestions if we can find another way to get this information.

sorc/chgres_cube.fd/surface.F90
JeffBeck-NOAA: @GeorgeGayno-NOAA, you can find the HRRR geogrid file on Hera here: /scratch2/BMC/det/beck/FV3-LAM/geo_em.d01.nc_HRRRX

JeffBeck-NOAA:
@GeorgeGayno-NOAA, @LarissaReames-NOAA added a description of all grib2-specific variables and put some short headers at the top of the two new subroutines to describe what they do.  The old subroutine that modified the landmask is also now commented out.  These were added in the last commit to the PR.

GeorgeGayno-NOAA:
I have a general question.  Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?

LarissaReames-NOAA:
> I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?

For that surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.

For the atmospheric component, if the *entire* grid lies outside of the input data, it also fails on the first RegridStore command. However, if only part of the grid lies outside of the input data, it fills the missing area with data from a nearby outermost data point.

GeorgeGayno-NOAA:
> > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> 
> For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> 
> For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> 
> I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?

Yes, we want it to fail for the atmosphere as well.  My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles.  But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore.  Anyway, we should experiment with the arguments sent to FieldRegridStore

JeffBeck-NOAA:
> > > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> > 
> > 
> > For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> > For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> > `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> > I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?
> 
> Yes, we want it to fail for the atmosphere as well. My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles. But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore. Anyway, we should experiment with the arguments sent to FieldRegridStore

@GeorgeGayno-NOAA, would you like @LarissaReames-NOAA and I to remove the call to EXTRAPMETHOD_NEAREST_STOD and then you can retest the global regression tests to ensure functionality isn't broken?

GeorgeGayno-NOAA:
> > > > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> > > 
> > > 
> > > For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> > > For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> > > `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> > > I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?
> > 
> > 
> > Yes, we want it to fail for the atmosphere as well. My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles. But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore. Anyway, we should experiment with the arguments sent to FieldRegridStore
> 
> @GeorgeGayno-NOAA, would you like @LarissaReames-NOAA and I to remove the call to EXTRAPMETHOD_NEAREST_STOD and then you can retest the global regression tests to ensure functionality isn't broken?

I removed that line from the head of 'develop' and started the regression tests on Hera.  Will let you know the results.  Do you have a canned case with a model grid outside the grib2 data grid?

sorc/chgres_cube.fd/input_data.F90
GeorgeGayno-NOAA: So wgrib2 will not retrieve information about the map projection (pass it back in memory)?  Maybe Wesley E. knows of a way to do that (I can ask him).  Also, the NCEP grib2 library will not only degrib, but can pass back all header information (the grid description template, etc).  That library is an option to retrieve the header info you need.

JeffBeck-NOAA:
> > > > > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> > > > 
> > > > 
> > > > For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> > > > For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> > > > `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> > > > I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?
> > > 
> > > 
> > > Yes, we want it to fail for the atmosphere as well. My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles. But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore. Anyway, we should experiment with the arguments sent to FieldRegridStore
> > 
> > 
> > @GeorgeGayno-NOAA, would you like @LarissaReames-NOAA and I to remove the call to EXTRAPMETHOD_NEAREST_STOD and then you can retest the global regression tests to ensure functionality isn't broken?
> 
> I removed that line from the head of 'develop' and started the regression tests on Hera. Will let you know the results. Do you have a canned case with a model grid outside the grib2 data grid?

@GeorgeGayno-NOAA, I removed the first line corresponding to that ESMF flag from atmosphere.F90, and after testing with a Alaska domain using CONUS external model data, I am now getting the correct error:

application called MPI_Abort(MPI_COMM_WORLD, 999) - process 1
 - FATAL ERROR: IN FieldRegridStore

From the PET files:

20200923 194413.726 ERROR            PET00 ESMCI_Mesh_Regrid_Glue.C:313 c_esmc_regrid_create() Arguments are incompatible  - - There exist destination points (e.g. id=1) which can't be mapped to any source cell

GeorgeGayno-NOAA:
> > > > > > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> > > > > 
> > > > > 
> > > > > For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> > > > > For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> > > > > `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> > > > > I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?
> > > > 
> > > > 
> > > > Yes, we want it to fail for the atmosphere as well. My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles. But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore. Anyway, we should experiment with the arguments sent to FieldRegridStore
> > > 
> > > 
> > > @GeorgeGayno-NOAA, would you like @LarissaReames-NOAA and I to remove the call to EXTRAPMETHOD_NEAREST_STOD and then you can retest the global regression tests to ensure functionality isn't broken?
> > 
> > 
> > I removed that line from the head of 'develop' and started the regression tests on Hera. Will let you know the results. Do you have a canned case with a model grid outside the grib2 data grid?
> 
> @GeorgeGayno-NOAA, I removed the first line corresponding to that ESMF flag from atmosphere.F90, and after testing with an Alaska domain using CONUS external model data, I am now getting the correct error:
> 
> application called MPI_Abort(MPI_COMM_WORLD, 999) - process 1
> FATAL ERROR: IN FieldRegridStore
> 
> From the PET files:
> 
> 20200923 194413.726 ERROR PET00 ESMCI_Mesh_Regrid_Glue.C:313 c_esmc_regrid_create() Arguments are incompatible - - There exist destination points (e.g. id=1) which can't be mapped to any source cell

That is great that it broke.  This error message should be part of the documentation.  All regression tests passed on Hera.   So removing the EXTRAPMETHOD flag does not appear to change results.

JeffBeck-NOAA:
> > > > > > > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> > > > > > 
> > > > > > 
> > > > > > For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> > > > > > For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> > > > > > `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> > > > > > I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?
> > > > > 
> > > > > 
> > > > > Yes, we want it to fail for the atmosphere as well. My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles. But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore. Anyway, we should experiment with the arguments sent to FieldRegridStore
> > > > 
> > > > 
> > > > @GeorgeGayno-NOAA, would you like @LarissaReames-NOAA and I to remove the call to EXTRAPMETHOD_NEAREST_STOD and then you can retest the global regression tests to ensure functionality isn't broken?
> > > 
> > > 
> > > I removed that line from the head of 'develop' and started the regression tests on Hera. Will let you know the results. Do you have a canned case with a model grid outside the grib2 data grid?
> > 
> > 
> > @GeorgeGayno-NOAA, I removed the first line corresponding to that ESMF flag from atmosphere.F90, and after testing with an Alaska domain using CONUS external model data, I am now getting the correct error:
> > application called MPI_Abort(MPI_COMM_WORLD, 999) - process 1
> > FATAL ERROR: IN FieldRegridStore
> > From the PET files:
> > 20200923 194413.726 ERROR PET00 ESMCI_Mesh_Regrid_Glue.C:313 c_esmc_regrid_create() Arguments are incompatible - - There exist destination points (e.g. id=1) which can't be mapped to any source cell
> 
> That is great that it broke. This error message should be part of the documentation. All regression tests passed on Hera. So removing the EXTRAPMETHOD flag does not appear to change results.

Glad to hear the regression tests passed too.  Would you like us to remove that flag in this PR and push a commit?

GeorgeGayno-NOAA:
> > > > > > > > I have a general question. Since these new GRIB2 data are regional, what happens if the model grid is partially or totally located outside the GRIB2 data grid?
> > > > > > > 
> > > > > > > 
> > > > > > > For the surface component, if any part of the target grid lies outside of the input data, it fails on the first RegridStore command.
> > > > > > > For the atmospheric component, it fills the missing area with data from a nearby outermost data point, even if the entire grid is outside of the input data. This behavior is the result of this flag:
> > > > > > > `extrapmethod=ESMF_EXTRAPMETHOD_NEAREST_STOD`
> > > > > > > I would think the ideal behavior would be for both codes to fail if the user attempts to use a target grid with any grid cell outside the input data, but we would need to change that flag to ESMF_EXTRAPMETHOD_NONE. What was the reasoning for including this flag in the global release?
> > > > > > 
> > > > > > 
> > > > > > Yes, we want it to fail for the atmosphere as well. My memory is hazy, but I believe the EXTRAPMETHOD_NEAREST_STOD was included to handle the poles. But now we use POLEMETHOD_ALLAVG, so it may not be needed anymore. Anyway, we should experiment with the arguments sent to FieldRegridStore
> > > > > 
> > > > > 
> > > > > @GeorgeGayno-NOAA, would you like @LarissaReames-NOAA and I to remove the call to EXTRAPMETHOD_NEAREST_STOD and then you can retest the global regression tests to ensure functionality isn't broken?
> > > > 
> > > > 
> > > > I removed that line from the head of 'develop' and started the regression tests on Hera. Will let you know the results. Do you have a canned case with a model grid outside the grib2 data grid?
> > > 
> > > 
> > > @GeorgeGayno-NOAA, I removed the first line corresponding to that ESMF flag from atmosphere.F90, and after testing with an Alaska domain using CONUS external model data, I am now getting the correct error:
> > > application called MPI_Abort(MPI_COMM_WORLD, 999) - process 1
> > > FATAL ERROR: IN FieldRegridStore
> > > From the PET files:
> > > 20200923 194413.726 ERROR PET00 ESMCI_Mesh_Regrid_Glue.C:313 c_esmc_regrid_create() Arguments are incompatible - - There exist destination points (e.g. id=1) which can't be mapped to any source cell
> > 
> > 
> > That is great that it broke. This error message should be part of the documentation. All regression tests passed on Hera. So removing the EXTRAPMETHOD flag does not appear to change results.
> 
> Glad to hear the regression tests passed too. Would you like us to remove that flag in this PR and push a commit?

Yes.

JeffBeck-NOAA:
@GeorgeGayno-NOAA Just pushed a commit that removes those ESMF flags.  Thanks, @LarissaReames-NOAA!

GeorgeGayno-NOAA:
@JeffBeck-NOAA  Your branch is missing stuff from 'develop', which was updated last week.  Can you do a merge to your branch?

sorc/chgres_cube.fd/grib2_util.F90
```diff
@@ -74,4 +74,171 @@ subroutine convert_omega(omega,p,t,q,clb,cub)
...
+          if (j>= iachar("a") .and. j<=iachar("z") ) then
+               strOut(i:i) = achar(iachar(strIn(i:i))-32)
+          else
+               strOut(i:i) = strIn(i:i)
+          end if
+     end do
+
+end function to_upper
+
+subroutine write_temp(filename,fieldname,localpet,field_input,field_target)
```
GeorgeGayno-NOAA, line 98: Is this routine used?

sorc/chgres_cube.fd/program_setup.f90
```diff
@@ -98,16 +94,71 @@ module program_setup
...
 ! vcoord_file_target_grid         Vertical coordinate definition file
 ! wltsmc_input/target             Wilting point soil moisture content
 !                                 input/target grids
-! 
+!
+! nsoill_out                      Number of soil levels desired in the output data. 
+!                                 chgres_cube can interpolate from 9 input to 4 output
+!                                 levels. 
+!                                 DEFAULT: 4    
+!
```
GeorgeGayno-NOAA, line 115: Thank you for adding this helpful documentation.

sorc/chgres_cube.fd/program_setup.f90
```diff
@@ -98,16 +94,71 @@ module program_setup
...
+!                                 can vary. 
+!                                 Default: False
+!
+! minmax_vgfrc_from_climo         If false, interpolate min/max vegetation fraction from 
+!                                 the input data to the target grid instead of using data
+!                                 from static data. Use with caution as vegetation
+!                                 categories can vary. 
+!                                 Default: False
+!
+! lai_from_climo                  If false, interpolate leaf area index from the input
```
GeorgeGayno-NOAA, line 152: FYI.  We have a global, weekly 5km LAI climo dataset available.  It can be mapped to any grid - regional or global - as part of the grid generation process.

sorc/chgres_cube.fd/program_setup.f90
```diff
@@ -98,16 +94,71 @@ module program_setup
...
+!                                 can vary. 
+!                                 Default: False
+!
+! sotyp_from_climo                If false, interpolate soil type from the input 
+!                                 data to the target grid instead of using data from 
+!                                 static data. Use with caution as the code assumes
+!                                 input soil type use STATSGO soil categories.
+!                                 Default: False
+!
+! vgfrc_from_climo                If false, interpolate vegetation fraction from the input
```
GeorgeGayno-NOAA, line 140: Why set veg fraction from the input file?  The grid generation process already creates that field from 0.144-degree monthly data.  Is that data too coarse?  If you have a better dataset, we should make it an option for the when making a model grid.   Also, the sfcsub.F component of the model will update veg fraction every 24-hours using either the gaussian file or the tiled file created during the grid generation process.  So, if you run with sfcsub.F turned on, your high-res veg fraction will get wiped out after one day.

sorc/chgres_cube.fd/program_setup.f90
```diff
@@ -98,16 +94,71 @@ module program_setup
...
+!                                 Assumes atmospheric and surface data are in a single 
+!                                 file. 
+!
+! varmap_file                     REQUIRED. Full path of the relevant varmap file. 
+!
+! external_model                  The model that the input data is derived from. Current
+!                                 supported options are: "GFS", "HRRR", "NAM", "RAP". 
+!                                 Default: "GFS"
+!
+! vgtyp_from_climo                If false, interpolate vegetation type from the input
```
GeorgeGayno-NOAA, line 128: We have high-res versions of the statsgo soil type data and igbp vegetation type available as part of the model grid generation process.  Do you require different datasets?  If so, we should consider making them available for all users.

sorc/chgres_cube.fd/program_setup.f90
```diff
@@ -293,6 +369,41 @@ subroutine read_setup_namelist
...
+ if (trim(input_type) == "grib2") then
+	 if (trim(grib2_file_input_grid) == "NULL" .or. trim(grib2_file_input_grid) == "") then
+		 call error_handler("FOR GRIB2 DATA, PLEASE PROVIDE GRIB2_FILE_INPUT_GRID", 1)
+	 endif
+ endif
+
+ !-------------------------------------------------------------------------
+! For grib2 input, warn about possibly unsupported external model types
+!-------------------------------------------------------------------------
+
```
GeorgeGayno-NOAA, line 386: I would stop processing here.  Other grib2 data will almost certainly not work.

sorc/chgres_cube.fd/search_util.f90
```diff
@@ -149,22 +166,39 @@ subroutine search (field, mask, idim, jdim, tile, field_num, latitude)
        elseif (field_num == 91) then  ! sea ice fract
          if (abs(latitude(i,j)) > 55.0) then
            field(i,j) = default_value
+           repl_default = repl_default + 1
          else
            field(i,j) = 0.0
+           repl_default = repl_default + 1
          endif
+       elseif (field_num == 7) then
```
GeorgeGayno-NOAA, line 174: Why does chgres need to adjust the terrain?  The 'oro' files provide the 'official' land mask and terrain.  chgres should map to it.

sorc/chgres_cube.fd/write_data.F90
```diff
@@ -2580,6 +2606,17 @@ subroutine write_fv3_sfc_data_netcdf(localpet)
      error = nf90_put_var( ncid, id_snoalb, dum2d, start=(/1,1,1/), count=(/i_target_out,j_target_out,1/))
      call netcdf_err(error, 'WRITING MAX SNOW ALBEDO RECORD' )
    endif
+
```
GeorgeGayno-NOAA, line 2609: If LAI is not being processed, what gets written to the file?  Missing values?  Does that record need to be written?

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -98,9 +103,13 @@ module input_data
...
  type(esmf_field), public        :: veg_type_input_grid     ! vegetation type
  type(esmf_field), public        :: z0_input_grid           ! roughness length
+ type(esmf_field), public        :: veg_greenness_input_grid ! vegetation fraction
+ type(esmf_field), public        :: lai_input_grid          ! leaf area index
+ type(esmf_field), public        :: max_veg_greenness_input_grid ! shdmax
+ type(esmf_field), public        :: min_veg_greenness_input_grid ! shdmin
 
- integer, public      :: lsoil_input=4  ! # of soil layers,
-                                                   ! # hardwire for now
+ integer, public      :: lsoil_input=4  ! # of soil layers, no longer hardwired to allow
```
GeorgeGayno-NOAA, line 111: Is this variable used anymore?

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -537,6 +546,42 @@ subroutine read_input_sfc_data(localpet)
                                    ungriddedUBound=(/lsoil_input/), rc=rc)
  if(ESMF_logFoundError(rcToCheck=rc,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
     call error_handler("IN FieldCreate", rc)
+    
+ 
+
+ if (.not. vgfrc_from_climo) then
+ print*,"- CALL FieldCreate FOR INPUT VEGETATION GREENNESS."
```
GeorgeGayno-NOAA, line 553: Be sure to add these new fields to 'cleanup_input_sfc_data'.

sorc/chgres_cube.fd/grib2_util.F90
LarissaReames-NOAA: No. I wrote that a while back to output data mid-exection for debugging purposes. I'll remove it from this branch.

JeffBeck-NOAA:
> @JeffBeck-NOAA Your branch is missing stuff from 'develop', which was updated last week. Can you do a merge to your branch?

@GeorgeGayno-NOAA Just merged develop into my branch.

sorc/chgres_cube.fd/search_util.f90
LarissaReames-NOAA: This only happens on the terrain_from_input_grid field. Despite the indication in this file that a flag value of  -99999.9 in the terrain height will turn off soil temperature adjustments for terrain height, that is not implemented in adjust_soilt_for_terrain. So, when using some of the new fields from HRRR data, I was getting isolated flag values at land ice points in input terrain height after interpolating it to the target grid, thus the soil temperature adjustment would fail. If we'd like to implement that flag value to turn off soil temperature adjustment, that would work instead.

sorc/chgres_cube.fd/input_data.F90
LarissaReames-NOAA: Yes, but it can now be changed by the code if it determines that the input data has 9 soil levels (e.g, HRRR data).

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA Yes, we use the 30 s BNU soil types from WRF and 15s MODIS vegetation types (also from WRF) in the HRRR.  Those are required for us to be able to directly compare FV3-LAM output to the operational HRRR to assess whether we're making progress with the RRFS toward implementation in 2023.

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA We use real-time NOAA VIIRS greenness fraction at 4-km resolution (this data is also in testing at MCEP) for the HRRR, and as I mentioned in the other comment, we need this to directly compare the FV3-LAM to the operational HRRR for readiness testing of the RRFS.  

We need to turn of the sfcsub.F component of the model for the FV3-LAM, based on what you're saying.  What are the namelist entries that control this part of the code?

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA We use 30 s MODIS LAI data available in WRF GEOG for our LAI in the HRRR, which is needed for our FV3-LAM to HRRR comparisons for RRFS readiness.  It would be good to provide these alternative, higher-resolution surface field options (for LAI, greenness/vegetation fraction, soil type, and vegetation type) in sfc_climo, then provide the user the option of which dataset they would like to use.

sorc/chgres_cube.fd/program_setup.f90
GeorgeGayno-NOAA: > @GeorgeGayno-NOAA We use real-time NOAA VIIRS greenness fraction at 4-km resolution (this data is also in testing at MCEP) for the HRRR, and as I mentioned in the other comment, we need this to directly compare the FV3-LAM to the operational HRRR for readiness testing of the RRFS.
> 
> We need to turn off the sfcsub.F component of the model for the FV3-LAM, based on what you're saying. What are the namelist entries that control this part of the code?

The vegetation fraction update may be turned off by setting its nudging coefficient as follows:  FVEGL=99999.

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA, @LarissaReames-NOAA and I wanted to provide the option for users who are advanced enough to try their own GRIB2 files, given that GRIB2 files are very standardized and have a better chance of working than other kinds of initialization datasets.  In addition, we provide the option for users to create their own varmap tables, so it would make sense that they could in turn try new data sources.  We do warn the user of the supported GRIB2 files to this point, so they should be aware that what they're trying may not work immediately.

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: Thanks, @GeorgeGayno-NOAA.  After some digging, @LarissaReames-NOAA and I ran into this issue previously and have set fhcyc in the FV3-LAM to avoid all surface nudging, since it isn't necessary at mesoscale NWP forecast lengths.

sorc/chgres_cube.fd/write_data.F90
LarissaReames-NOAA: I've now added in logic that will only create and write to these output fields if lai_from_climo is true.

sorc/chgres_cube.fd/search_util.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA I've updated the PR to include code changes from the other comments in this thread.  If we want to turn off soil temperature adjustment over these isolated points, that would be the last things to change.  Thanks.

sorc/chgres_cube.fd/input_data.F90
LarissaReames-NOAA: They've been added. I also added an additional field to the corresponding routine for target fields for now. It's related to the terrain height adjustment, so it may be removed depending on how we decided to proceed with that flag.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -164,26 +170,54 @@ subroutine surface_driver(localpet)
...
+ 
 !---------------------------------------------------------------------------------------------
 ! Rescale soil moisture for changes in soil type between the input and target grids.
 !---------------------------------------------------------------------------------------------
 
  call rescale_soil_moisture
+ 
+ if (.not. (vgtyp_from_climo .or. sotyp_from_climo)) then
+  !---------------------------------------------------------------------------------------------
+  ! Check soil moisture again for mismatches after rescale_soil_moisture subroutine
```
GeorgeGayno-NOAA, line 210: Why are there mismatches?  Do they first appear after the call to 'rescale_soil_moisture', or earlier?

sorc/chgres_cube.fd/surface.F90
```diff
@@ -164,26 +170,54 @@ subroutine surface_driver(localpet)
...
+! Adjust soil/landice column temperatures for any change in elevation between
+! the
 ! input and target grids.
 !---------------------------------------------------------------------------------------------
 
  call adjust_soilt_for_terrain
-
+ 
 !---------------------------------------------------------------------------------------------
 ! Rescale soil moisture for changes in soil type between the input and target grids.
```
GeorgeGayno-NOAA, line 203: FYI, The soil moisture rescaling assumes Noah LSM land states on input and output.  Does the HRRR data contain RUC LSM land states?  If so, this routine is not relevant.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -2234,6 +2609,518 @@ subroutine calc_liq_soil_moisture
```
GeorgeGayno-NOAA, line 2609: FYI.  calc_liq_soil_moisture was written to support the Noah-LSM.  It is likely not relevant for RUC-LSM.  I know nothing about land models other than Noah.   So expanding chgres_cube to work with input data sets with land states from other land models will need to be carefully done.

sorc/chgres_cube.fd/surface.F90
LarissaReames-NOAA: HRRR data is indeed RUC LSM data, but the soil types are the same classification system so I assumed this would still be relevant. However, I do believe the RUC LSM states (particularly how RUC treats soil type under land ice) are the cause of the soil moisture mismatches that you mentioned in a prior question.

sorc/chgres_cube.fd/surface.F90
LarissaReames-NOAA: HRRR data is indeed RUC LSM data, but the soil types are the same classification system so I assumed this would still be relevant. However, I do believe the RUC LSM states (particularly how RUC treats soil type under land ice) are the cause of the soil moisture mismatches that you mentioned in a prior question.

sorc/chgres_cube.fd/program_setup.f90
GeorgeGayno-NOAA: GRIB2 is a standard data format, but the contents of a GRIB2 file is certainly not standard.  The surface component of chgres_cube is very Noah-LSM-centric.  Ingesting land states from models other than Noah will not work well.  A novice user may get chgres to run without error with his GRIB2 data, but the initial land states may not be what he expects.  Land states require a lot of special handling.  It is more involved than, say, using 500 mb temperature from other dynamic core.

LarissaReames-NOAA: Could you point us to the places in the code that are Noah-LSM centric? I'm not quite sure that part is clear to me. I might be able to provide a better response once I understand that distinction.

JeffBeck-NOAA: @GeorgeGayno-NOAA My understanding from discussions with @LarissaReames-NOAA is that the soil and vegetation classifications for RUC-LSM and NOAH LSM are identical.  We've tested both RAP and HRRR data (that use RUC-LSM) extensively with this new code and are confident that we're providing FV3 with the correct soil levels/values, etc.  However, as you mention, it is possible that a user may try a new, unknown LSM that would be incompatible or may not provide results that are expected.  What if we cause a failure for all unknown GRIB2 source data, but then tell users that they may try to use the data **at their own risk** by modifying the code?

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -49,6 +54,8 @@ module input_data
                                     longitude_input_grid, &
                                     inv_file
 
+ use surface
```
kgerheiser, line 57: There's a circular dependency between `surface.F90` and `input_data.F90` that causes the CMake build to fail for me.

LarissaReames-NOAA: @kgerheiser Thanks for pointing that out. Looks like that line got left hanging there on accident. I'll remove it.

GeorgeGayno-NOAA:
Note the bug fix that will be added to the medium weather release branch #155.  The wgrib2 inventory file (in model_grid.F90) is being written by all mpi tasks, which caused problems with large task counts.  I will add this bug fix to 'develop'.  Check your branch for potential problems as well.

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA The latest commit pushed includes a change to fail if the user chooses an unsupported grib2 file, which instructions to proceed at their own risk, if they wish.

JeffBeck-NOAA:
> Note the bug fix that will be added to the medium weather release branch #155. The wgrib2 inventory file (in model_grid.F90) is being written by all mpi tasks, which caused problems with large task counts. I will add this bug fix to 'develop'. Check your branch for potential problems as well.

@GeorgeGayno-NOAA The latest commit to this PR now includes this bug fix.  Thanks @LarissaReames-NOAA!

sorc/chgres_cube.fd/model_grid.F90
```diff
@@ -781,12 +795,331 @@ subroutine define_input_grid_gfs_grib2(localpet, npets)
...
+     error=nf90_inq_varid(ncid, 'gridlat', id_var)
+     call netcdf_err(error, 'reading field id' )
+     error=nf90_get_var(ncid, id_var, latitude_one_tile)
+     call netcdf_err(error, 'reading field' )
+
+     error=nf90_inq_varid(ncid, 'gridlon', id_var)
+     call netcdf_err(error, 'reading field id' )
+     error=nf90_get_var(ncid, id_var, longitude_one_tile)
+     call netcdf_err(error, 'reading field' )
+
```
GeorgeGayno-NOAA, line 885: Is this instance of grb2_mk_inv being called on all tasks?  This is instance and the one at line 897 be moved before the 'if' statement?

JeffBeck-NOAA: Thanks, @GeorgeGayno-NOAA.  This instance is now behind an appropriate if statement to run as a single task.  I just pushed the commit.

JeffBeck-NOAA: @GeorgeGayno-NOAA, the print statement was also moved into the "if" statement too.

sorc/chgres_cube.fd/program_setup.f90
GeorgeGayno-NOAA: > Could you point us to the places in the code that are Noah-LSM centric? I'm not quite sure that part is clear to me. I might be able to provide a better response once I understand that distinction.

Here are the EMC Noah-LSM-centric parts of the code:

Routine 'calc_soil_params_driver' computes soil thresholds, such as wilting point, according to the soil type and the soil hydraulic coefficients.  The coefficients are those use by EMC's Noah-lsm.  I can't say what thresholds other LSMs use.

The above soil thresholds are used in 'rescale_soil_moisture' to rescale the soil moisture based on differences between the input and target grids.  The method is specifically implemented for the Noah-lsm.  Is it applicable to other LSMs?  I don't know.

The function FHR20 computes the frozen portion of the total soil moisture.  This function is part of the Noah-lsm. I don't know what other land models use to compute frozen soil.

sorc/chgres_cube.fd/model_grid.F90
GeorgeGayno-NOAA: Sorry, I meant to ask if the instances at 886 and 899 could be combined and moved before the 'if' at 856.

LarissaReames-NOAA: If we have it create an inventory file before checking the template number, it's possible the inventory file could be created for a template number that won't be compatible with the code, and then it would fail. I didn't think that was very efficient.

GeorgeGayno-NOAA: Ok.  That answers my question.

JeffBeck-NOAA:
@GeorgeGayno-NOAA @LarissaReames-NOAA Latest commit fixes the extrapolation failure for winds.

sorc/chgres_cube.fd/program_setup.f90
JeffBeck-NOAA: @GeorgeGayno-NOAA @LarissaReames-NOAA, I have not heard back from Tanya regarding whether the equations used in these routines are equivalent to what RUC-LSM does or whether we will need to incorporate different calculations for RUC-LSM vs NOAH based on the number of soil levels requested (9 for RUC-LSM, 4 for NOAH).  Given the code slush on Wednesday, Sept. 30, can we merge this PR and create a UFS_UTILS release/public-v2 branch now?  Then we can open a PR to both develop and release/public-v2 if we have these RUC-LSM changes to merge later.  We're up against this deadline to make sure the code is ready to start porting to other platforms on which we've not yet tested, and it would be good to get that started ASAP.  Thanks.

GeorgeGayno-NOAA:
Tried to compile 34fb1ef on WCOSS-Cray.  It could not find the compilers.  Very odd, because it seems like all the required (including PrgEnv) modules were loaded.  The only difference is loading the build module instead of sourcing it.  The Cray can be an odd machine.

```
-- The C compiler identification is unknown
-- The Fortran compiler identification is unknown
Error: A PrgEnv-* modulefile must be loaded.-- Check for working C compiler: /opt/cray/craype/2.3.0/bin/cc
-- Check for working C compiler: /opt/cray/craype/2.3.0/bin/cc -- broken
CMake Error at /gpfs/hps/usrx/local/prod/cmake/3.16.2/share/cmake-3.16/Modules/CMakeTestCCompiler.cmake:60 (message):
  The C compiler

    "/opt/cray/craype/2.3.0/bin/cc"

  is not able to compile a simple test program.```

Won't have time to investigate further today.

**Comment by ywangwof** [comment]
> Tried to compile [34fb1ef](https://github.com/NOAA-EMC/UFS_UTILS/commit/34fb1ef7ffc95784b835916d85f77cf85efb2e26) on WCOSS-Cray. It could not find the compilers. Very odd, because it seems like all the required (including PrgEnv) modules were loaded. The only difference is loading the build module instead of sourcing it. The Cray can be an odd machine. Won't have time to investigate further today.
> 
> ```
> -- The C compiler identification is unknown
> -- The Fortran compiler identification is unknown
> Error: A PrgEnv-* modulefile must be loaded.-- Check for working C compiler: /opt/cray/craype/2.3.0/bin/cc
> -- Check for working C compiler: /opt/cray/craype/2.3.0/bin/cc -- broken
> CMake Error at /gpfs/hps/usrx/local/prod/cmake/3.16.2/share/cmake-3.16/Modules/CMakeTestCCompiler.cmake:60 (message):
>   The C compiler
> 
>     "/opt/cray/craype/2.3.0/bin/cc"
> 
>   is not able to compile a simple test program.```
> ```

On Odin, I have the same problem. It is shown that I cannot use 
`module purge`
but I have to use
`module swap module_name/version`
to replace _module_name_ with a specific version. Or just use `module load module_name` to load the same module again and it is not an error. Furthermore, once a module with command _module load_ (for example, _modulefiles/odin.intel/fv3_) is loaded, it cannot be unloaded again.

Yunheng.

**Comment by kgerheiser** [comment]
In the build script, when you do this `> /dev/null 2>&1` it hides a bug related to loading ESMF that somehow propagates into compiler not found, for whatever reason.

If you try to load to load `esmf/8.0.0` it gives an error esmf/8.0.0(9):ERROR:151: Module 'esmf/8.0.0' depends on one of the module(s) 'intel/16.3.210' If you comment out module load esmf/8.0.0  (ESMFMKFILE is set which is all it needs) everything works.

**Comment by GeorgeGayno-NOAA** [comment]
> In the build script, when you do this `> /dev/null 2>&1` it hides a bug related to loading ESMF that somehow propagates into compiler not found, for whatever reason.
> 
> If you try to load to load `esmf/8.0.0` it gives an error `esmf/8.0.0(9):ERROR:151: Module 'esmf/8.0.0' depends on one of the module(s) 'intel/16.3.210'`
> 
> If you comment out or remove the line `module load esmf/8.0.0` (ESMFMKFILE is set which is all it needs) everything works.

It worked for me too.  Thanks.

**Comment by JeffBeck-NOAA** [comment]
> In the build script, when you do this `> /dev/null 2>&1` it hides a bug related to loading ESMF that somehow propagates into compiler not found, for whatever reason.
> 
> If you try to load to load `esmf/8.0.0` it gives an error `esmf/8.0.0(9):ERROR:151: Module 'esmf/8.0.0' depends on one of the module(s) 'intel/16.3.210'`
> 
> If you comment out or remove the line `module load esmf/8.0.0` (ESMFMKFILE is set which is all it needs) everything works.

@GeorgeGayno-NOAA, @kgerheiser, @LarissaReames-NOAA, Thanks to Kyle for helping to find this bug.  The latest commit comments out the "module load esmf/8.0.0" line in the WCOSS-Cray module file.

File: sorc/chgres_cube.fd/model_grid.F90 [hunk 22]
```diff
@@ -781,12 +795,334 @@ subroutine define_input_grid_gfs_grib2(localpet, npets)
...
+ if(ESMF_logFoundError(rcToCheck=error,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
+      call error_handler("IN FieldGather", error)
+
+
+  ! Check for the grid template number to see if wgrib2 can access the lat/lon arrays
+  if (localpet==0) then
+    cmdline_msg = "wgrib2 "//trim(the_file)//" -d 1 -Sec3 &> temp.out"
+    call system(cmdline_msg)
+  endif
+  call MPI_BARRIER(MPI_COMM_WORLD, error)
```
GeorgeGayno-NOAA, line 844: I spoke with Wesley E. about how to return the template info in memory.  He said you can call grb2_inq with argument 'grid_desc=' to return it as a text string.  More info is here: https://www.cpc.ncep.noaa.gov/products/wesley/wgrib2/grb2_inq.html.  If that works, you won't have to carry around the wgrib2 executable.

build_all.sh
```diff
@@ -3,21 +3,22 @@ set -eux
 
 target=${target:-"NULL"}
 
-if [[ "$target" == "linux.gnu" || "$target" == "linux.intel" ]]; then
+export MOD_PATH
+
+if [[ "$target" == "linux" || "$target" == "macosx" ]]; then
```
climbfuji, line 8: do you still need the .intel, .gnu suffices for both linux and macos? If so, use
```
if [[ "$target" == "linux.*" || "$target" == "macosx.*" ]]; then
```
(I think that works, but someone needs to test)

JeffBeck-NOAA:
If we plan to support both Intel and gnu compilers on Linux and Macosx platforms, then I can add that in. Since ufs-weather-model only supports gnu on those systems (right?), I didn't include it.

climbfuji:
> If we plan to support both Intel and gnu compilers on Linux and Macosx platforms, then I can add that in. Since ufs-weather-model only supports gnu on those systems (right?), I didn't include it.
> […](#)

In the ufs-weather-model, we have explicit build targets linux.gnu, linux.intel, macosx.gnu (but not yet macosx.intel). I suggestt to follow this nomenclature for consistency, even if only the gnu version is used at the moment.

JeffBeck-NOAA:
@DomHeinzeller I just pushed a commit to add the wildcards for linux/macosx gnu/intel compilers.

sorc/chgres_cube.fd/model_grid.F90
LarissaReames-NOAA: The latest push to my fork contains a commit that replaces this use of the wgrib2 executable with code that use grb2_inq instead.

GeorgeGayno-NOAA: Great.  That is less 'clunky'.

LarissaReames-NOAA:
> I see several new routines in surface.F90, such as "replace_land_sfcparams", "check_smois_land", "check_smois_water". What are these routines correcting?

These routines are now no longer used. The source of their creation was a mismatch between the soil type being read in from the geogrid file for HRRR and RAP data and the landmask that exists in the grib2 files. This mismatch also exists with the landmask in the geogrid file itself. To address this issue, after the soil type is read in from HRRR or RAP data, input grid points where the soil indicates water (14) but the input landmask is land, the next most common soil type (as indicated by the array of soil type fractions also pulled from geogrid) is used at that point instead. This correction is done in read_input_sfc_grib2_file to avoid having to do anything after interpolation to the target grid. The code for routines in question have been maintained just incase, but they are no longer used as the calls to them in surface.F90 have been commented out.

GeorgeGayno-NOAA:
> > I see several new routines in surface.F90, such as "replace_land_sfcparams", "check_smois_land", "check_smois_water". What are these routines correcting?
> 
> @GeorgeGayno-NOAA These routines are now no longer used. The source of their creation was a mismatch between the soil type being read in from the geogrid file for HRRR and RAP data and the landmask that exists in the grib2 files. This mismatch also exists with the landmask in the geogrid file itself. To address this issue, after the soil type is read in from HRRR or RAP data, input grid points where the soil indicates water (14) but the input landmask is land, the next most common soil type (as indicated by the array of soil type fractions also pulled from geogrid) is used at that point instead. This correction is done in read_input_sfc_grib2_file to avoid having to do anything after interpolation to the target grid. The code for the routines in question have been maintained just in case, but they are no longer used as the calls to them in surface.F90 have been commented out. Also note this hasn't been merged in to the PR yet as Jeff is off this week, but my branch is up to date with his and has these changes.

Glad to hear these routines are no longer used.  QCing the data in the read routine is much safer than passing mismatched data to surface.F90.  The code should be much more robust now.

GeorgeGayno-NOAA:
Took a quick look at Jeff's branch.  A recent update to develop (745fdbd) replaced instances of 'mpif.h' with 'use mpi" and added an 'mpi_barrier' after calls to 'grb_mk_inv'.  I saw one 'mpif.h'.  And one call to 'grb_mk_in'v did not appear to have an 'mpi_barrier' after it (see line 888 of model_grid.F90).   Please double check.  That 'grb_mk_inv' error was very insidious.

GeorgeGayno-NOAA:
@JeffBeck-NOAA Under ./CMakeModules, your branch is pointing to 09ff76f.  The authoritative 'develop' is pointing to a later version - 3eaddb8.  When you merge from 'develop', are you getting an alert?

LarissaReames-NOAA:
> Took a quick look at Jeff's branch. A recent update to develop ([745fdbd](https://github.com/NOAA-EMC/UFS_UTILS/commit/745fdbdee648ef8426bd069638db374f0affb1f9)) replaced instances of 'mpif.h' with 'use mpi" and added an 'mpi_barrier' after calls to 'grb_mk_inv'. I saw one 'mpif.h'. And one call to 'grb_mk_in'v did not appear to have an 'mpi_barrier' after it (see line 888 of model_grid.F90). Please double check. That 'grb_mk_inv' error was very insidious.

That line in Jeff's fork is now outdated with my updates to remove reliance on the system calls to the wgrib2 executable. I'll remove the mpif.h reference in that file in favor of use mpi.

GeorgeGayno-NOAA:
A note concerning the build modules.  Changing them to be true modules, then doing a module load in the build script is perfectly fine.  However, be aware those build modules are still being 'sourced' from the utility and regression test scripts.  When I tried to create a C96 grid using ./driver_scripts/driver_grid_dell.sh, I got a warning about 'setenv': command not found.  So, I made this change and the warning went away.  

diff --git a/driver_scripts/driver_grid.dell.sh b/driver_scripts/driver_grid.dell.sh
index d2d6d08d..760c4415 100755
--- a/driver_scripts/driver_grid.dell.sh
+++ b/driver_scripts/driver_grid.dell.sh
@@ -59,7 +59,8 @@
 #-----------------------------------------------------------------------

 source ../sorc/machine-setup.sh > /dev/null 2>&1
-source ../modulefiles/build.$target
+module use ../modulefiles
+module load build.$target
 module list

I can identify the scripts that needed to be updated and test them.

GeorgeGayno-NOAA:
> A note concerning the build modules. Changing them to be true modules, then doing a module load in the build script is perfectly fine. However, be aware those build modules are still being 'sourced' from the utility and regression test scripts. When I tried to create a C96 grid using ./driver_scripts/driver_grid_dell.sh, I got a warning about 'setenv': command not found. So, I made this change and the warning went away.
> 
> diff --git a/driver_scripts/driver_grid.dell.sh b/driver_scripts/driver_grid.dell.sh
> index d2d6d08d..760c4415 100755
> --- a/driver_scripts/driver_grid.dell.sh
> +++ b/driver_scripts/driver_grid.dell.sh
> @@ -59,7 +59,8 @@
> #-----------------------------------------------------------------------
> 
> source ../sorc/machine-setup.sh > /dev/null 2>&1
> -source ../modulefiles/build.$target
> +module use ../modulefiles
> +module load build.$target
> module list
> 
> I can identify the scripts that needed to be updated and test them.

I identified the changes required for WCOSS-Dell.  I don't want to update your branch.  Instead, I will send the 'git diff' to a file.  You can then apply the file to your branch using 'git apply gitdiff.file'.  Will work on WCOSS-Cray next.

JeffBeck-NOAA:
> @JeffBeck-NOAA Under ./CMakeModules, your branch is pointing to 09ff76f. The authoritative 'develop' is pointing to a later version - 3eaddb8. When you merge from 'develop', are you getting an alert?

@GeorgeGayno-NOAA, sorry for the late reply.  I was out this week on vacation.  I pulled the latest develop branch code into my fork/branch, and it now points to 3eaddb8.

JeffBeck-NOAA:
@GeorgeGayno-NOAA @LarissaReames-NOAA, the UFS_UTILS regression tests have passed for this PR on both Jet and Hera here:

Jet: /lfs4/BMC/wrfruc/beck/UFS_UTILS_Jeff/reg_tests/chgres_cube/regression.log
Hera: /scratch2/BMC/det/beck/FV3-LAM/UFS_UTILS_Jeff/reg_tests/chgres_cube/regression.log

This leaves us with Orion and WCOSS Dell and Cray to test.

sorc/chgres_cube.fd/model_grid.F90
```diff
@@ -788,12 +796,317 @@ subroutine define_input_grid_gfs_grib2(localpet, npets)
...
+
+ print*,"- I/J DIMENSIONS OF THE INPUT GRID TILES ", i_input, j_input
+
+ ip1_input = i_input + 1
+ jp1_input = j_input + 1
+
+ if (mod(npets,num_tiles_input_grid) /= 0) then
+   call error_handler("MUST RUN WITH A TASK COUNT THAT IS A MULTIPLE OF 6.", 1)
+ endif
+
```
GeorgeGayno-NOAA, line 914: In this case, the number of input tiles will always be one.  So this error message will never be tripped.

JeffBeck-NOAA:
> > > A note concerning the build modules. Changing them to be true modules, then doing a module load in the build script is perfectly fine. However, be aware those build modules are still being 'sourced' from the utility and regression test scripts. When I tried to create a C96 grid using ./driver_scripts/driver_grid_dell.sh, I got a warning about 'setenv': command not found. So, I made this change and the warning went away.
> > > diff --git a/driver_scripts/driver_grid.dell.sh b/driver_scripts/driver_grid.dell.sh
> > > index d2d6d08d..760c4415 100755
> > > --- a/driver_scripts/driver_grid.dell.sh
> > > +++ b/driver_scripts/driver_grid.dell.sh
> > > @@ -59,7 +59,8 @@
> > > #-----------------------------------------------------------------------
> > > source ../sorc/machine-setup.sh > /dev/null 2>&1
> > > -source ../modulefiles/build.$target
> > > +module use ../modulefiles
> > > +module load build.$target
> > > module list
> > > I can identify the scripts that needed to be updated and test them.
> > 
> > 
> > @GeorgeGayno-NOAA, sure! Please point me to the 'gitdiff' file and I'll be happy to incorporate those changes.
> 
> Find the 'git diff' files on Hera: /scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/UFS_UTILS.jeffb/$machine.diff.txt.

Hi @GeorgeGayno-NOAA, I just pushed a commit to add the diff files (including a minor tweak to make sure the c96 regional regression test succeeds, since the baseline files are still "tile1" instead of "tile7").

sorc/chgres_cube.fd/model_grid.F90
JeffBeck-NOAA: @GeorgeGayno-NOAA, I merged in Larissa's change to remove this check in the latest commit below.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -96,6 +98,9 @@ module surface
  type(esmf_field)                   :: terrain_from_input_grid
                                        ! terrain height interpolated
                                        ! from input grid
+ type(esmf_field)                   :: terrain_from_input_grid_land
+                                       ! terrain height interpolated
+                                       ! from input grid at all land points
```
GeorgeGayno-NOAA, line 104: How is terrain_from_input_grid_land used?  It is computed below.  But I don't understand how it is used.

LarissaReames-NOAA: We were experiencing issues where isolated land ice points weren't ever able to obtain a proper terrain height, so this array is passed through during the search/replace of land ice points. Since soil temperature adjustment for terrain height is not turned off when points on the target grid have the default terrain height replacement value (default_value = -99999.9), providing data interpolated to "land points" which the input data originally deemed these land ice points to be (since it doesn't have land ice) allows for these points to have a proper terrain height. It looks like this use actually got coded out, but I'm putting it back in by providing the data from terrain_from_input_grid as an optional argument to the search function on line 2049 of surface.F90.

BenjaminBlake-NOAA:
All of the regression tests for chgres cube passed on the WCOSS Dell after I added Jeff's tile1/tile7 fix to the c96 regional regression test script.  I'm waiting on the results from the WCOSS Cray but I believe they will also be successful.

sorc/chgres_cube.fd/model_grid.F90
```diff
@@ -788,12 +796,313 @@ subroutine define_input_grid_gfs_grib2(localpet, npets)
...
+             lat_src_ptr(i,j) = latitude_one_tile(1,j) - (0.5_esmf_kind_r8*deltalon)
+           endif
+
+         enddo
+       enddo
+     else
+       if (localpet==0) then
+         cmdline_msg = "wgrib2 "//trim(the_file)//" -d 1 -grid &> temp2.out"
+         call system(cmdline_msg)
+         open(4,file="temp2.out")
```
GeorgeGayno-NOAA, line 1046: I thought this was replaced by a call to grb2_inq.

LarissaReames-NOAA: Ah, it looks like you found another use that I had forgotten about. Thanks for pointing that out. I should able to replace that relatively quickly.

BenjaminBlake-NOAA:
The chgres cube regression tests all passed on the WCOSS Cray as well.

JeffBeck-NOAA:
@GeorgeGayno-NOAA I just merged in @LarissaReames-NOAA's latest code fixes for the terrain and wgrib2 change.

GeorgeGayno-NOAA:
There is something wrong with your latest commit (3393229).  I see two 'mpif.h' in model_grid.F90.  I thought they were all removed.  Why are they reappearing?  Also, an error message removed by Larissa has reappeared.  We need to stop these fork-to-fork merges.  The only merges should be from/to 'develop' of the authoritative repo.

LarissaReames-NOAA:
> There is something wrong with your latest commit ([3393229](https://github.com/NOAA-EMC/UFS_UTILS/commit/3393229c347bebfa40ab0b28e73625d10fbe46af)). I see two 'mpif.h' in model_grid.F90. I thought they were all removed. Why are they reappearing? Also, an error message removed by Larissa has reappeared. We need to stop these fork-to-fork merges. The only merges should be from/to 'develop' of the authoritative repo.

I'll now be merging straight to Jeff's branch. Those overwritten changes have now been put back in.

reg_tests/chgres_cube/driver.hera.sh_Jeff
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
+
```
GeorgeGayno-NOAA, line 2: This version should be removed.

JeffBeck-NOAA: @GeorgeGayno-NOAA Just removed that script.

JeffBeck-NOAA:
@GeorgeGayno-NOAA, the develop branch of UFS_UTILS fails all regression tests on Orion with very small (1e-10 to 1e-17 differences), likely related to the baseline files either having been created on a different platform, or by using different compilers.  However, I used the develop branch regression test results as the baseline for a test with the new code, and when I do that, the regression tests pass on Orion for this PR.

Therefore, all platforms (Orion, Hera, Jet, WCOSS Dell and Cray) now pass the regression tests for this PR.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA, the develop branch of UFS_UTILS fails all regression tests on Orion with very small (1e-10 to 1e-17 differences), likely related to the baseline files either having been created on a different platform, or by using different compilers. However, I used the develop branch regression test results as the baseline for a test with the new code, and when I do that, the regression tests pass on Orion for this PR.
> 
> Therefore, all platforms (Orion, Hera, Jet, WCOSS Dell and Cray) now pass the regression tests for this PR.

Ok.  I will do the merge today.

PR #156, opened 2020-09-25
Make size of mosaic_name string in make_solo_mosaic longer

gsketefian:
Change the length of the string mosaic_name in make_solo_mosaic.c from just 128 characters to STRING=255 characters.

This is needed because the way make_solo_mosaic is called in regional_workflow, mosaic_name includes not only the basename of the mosaic file but also the absolute path to it, and that path can get longer than 128 characters.  In that case, make_solo_mosaic fails (without a clear error message).  It would be even better to make mosaic_name a variable length string, but I don't remember how to do that in c.  This suffices for now.

I tested this with ~25 different tests in regional_workflow with varying mosaic_name lengths, and they all passed (whereas some of them failed previously with only a 128 character mosaic_name).


GeorgeGayno-NOAA:
Checked out his branch at 13799b7.  Successfully compiled it on Hera, Jet, Orion, WCOSS-Cray and WCOSS-Dell.  The grid generation regression tests also passed.  Will merge.

GeorgeGayno-NOAA (approved):
I approve.

PR #189, opened 2020-10-27
release/public-v2: update to build on gaea

climbfuji:
This PR is for the SRW App release 1.0, see also https://github.com/ufs-community/ufs-srweather-app/pull/43 and PRs listed there.

Enable UFS_UTILS build on gaea with minor modifications of `CMakeLists.txt`, `build_all.sh`, `sorc/machine-setup.sh`, and the addition of `modulefiles/build.gaea.intel` (module file identical to what will be committed to the ufs-weather-model in PR https://github.com/ufs-community/ufs-weather-model/pull/248).

OUTDATED: For some reason, this PR shows more commits than just mine when choosing release/public-v2 as target branch. It is still the branch with the fewest commits and smallest number of changes. I got my source code directory from @JeffBeck-NOAA so he should know what is going on.

With these changes, I was able to build UFS_UTILS on gaea successfully.


JeffBeck-NOAA:
@DomHeinzeller, the reason you see more commits than just yours is because I hadn't changed the Externals.cfg file to clone the release/public-v2 branch of UFS_UTILS instead of the develop branch.  So your PR also includes recent changes to the develop branch that we don't have in release/public-v2.  I don't believe this will be a problem, as all the other commits look good to me.  I'm tagging @LarissaReames-NOAA so that she's aware.  

The only major difference that this will introduce is the new modulesfiles that also have the ".compiler" extension.  This change is (at the moment) incompatible with the regional_workflow code that sources the UFS_UTILS module file based on what platform you're on.  However, if I understand the proposed build changes, we *should* be able to source only the NCEPLIBS module file and it will work across all executables from UFS_UTILS, ufs-weather-model, and EMC_post.  Is that correct?

climbfuji:
> @DomHeinzeller, the reason you see more commits than just yours is because I hadn't changed the Externals.cfg file to clone the release/public-v2 branch of UFS_UTILS instead of the develop branch. So your PR also includes recent changes to the develop branch that we don't have in release/public-v2. I don't believe this will be a problem, as all the other commits look good to me. I'm tagging @LarissaReames-NOAA so that she's aware.
> 
> The only major difference that this will introduce is the new modules files that also have the ".compiler" extension. This change is (at the moment) incompatible with the regional_workflow code that sources the UFS_UTILS module file based on what platform you're on. However, if I understand the proposed build changes, we _should_ be able to source only the NCEPLIBS module file and it will work across all executables from UFS_UTILS, ufs-weather-model, and EMC_post. Is that correct?

I think so, yes. I am not happy to have modulefiles in each ufs-weather-model, UFS_UTILS and EMC_post. Such an easy way to make mistakes, for example when updating one but forgetting about (one of) the others. But I guess there is no better solution if each of them should build standalone, too.

edwardhartnett:
Please be careful. We are starting to make changes on the develop branch which would impact public-v2. Don't grab any changes from the develop branch without careful examination and thought.

If you can easily re-create this PR without any changes from develop, I think that would be wise. As far as I recall, the only change so far has been to the README.

climbfuji:
> Please be careful. We are starting to make changes on the develop branch which would impact public-v2. Don't grab any changes from the develop branch without careful examination and thought.
> 
> If you can easily re-create this PR without any changes from develop, I think that would be wise. As far as I recall, the only change so far has been to the README.

Agreed. I recreated the PR based on a fresh clone of release/public-v2. @JeffBeck-NOAA @LarissaReames-NOAA @GeorgeGayno-NOAA please have a look, in particular check the additional "if intel" condition in `build_all.sh` (that came from the develop version of the file). Thanks.

build_all.sh
```diff
@@ -28,8 +28,10 @@ cd ./build
 
 CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=../ -DEMC_EXEC_DIR=ON"
 
-if [[ "$target" != "wcoss_cray" && "$target" != "odin" ]]; then
-  CMAKE_FLAGS+=" -DCMAKE_Fortran_COMPILER=ifort -DCMAKE_C_COMPILER=icc"
+if [[ "$compiler" == "intel" ]]; then
```
climbfuji, line 31: Why is this entire block necessary at all? The compilers should be set correctly by loading the modules, not by adding duct tape inside the build scripts.

GeorgeGayno-NOAA: So, in the build module do: setenv CMAKE_Fortran_COMPILER ifort

climbfuji: For the sake of this PR moving ahead, shall we just keep it as it is?

GeorgeGayno-NOAA: I can remove this block and test the compile on Hera, Orion, Jet, WCOSS-Cray, WCOSS-Dell.

GeorgeGayno-NOAA:
The release branch in the authoritative repo has moved ahead of your version.  You may need to do a merge.

climbfuji:
> The release branch in the authoritative repo has moved ahead of your version. You may need to do a merge.

Should be up to date now, thanks for letting me know.

build_all.sh
GeorgeGayno-NOAA: If I remove the entire block, the build will default to the GNU compiler (on the machines I have access to).  So additional changes will be necessary.  But if I just remove the first 'if' block ("$compiler == "intel") it works correctly.

climbfuji: > If I remove the entire block, the build will default to the GNU compiler (on the machines I have access to). So additional changes will be necessary. But if I just remove the first 'if' block ("$compiler == "intel") it works correctly.

Thanks for trying. We need to keep the outer intel test until we fix the cmake build, otherwise it would break the build for someone trying to use the GNU compilers on one of these platforms. I think the PR is good to go in its current form.

GeorgeGayno-NOAA: > > If I remove the entire block, the build will default to the GNU compiler (on the machines I have access to). So additional changes will be necessary. But if I just remove the first 'if' block ("$compiler == "intel") it works correctly.
> 
> Thanks for trying. We need to keep the outer intel test until we fix the cmake build, otherwise it would break the build for someone trying to use the GNU compilers on one of these platforms. I think the PR is good to go in its current form.

If the outer 'if' block is kept, I might add something like "compiler=${compiler:-intel}".  I did not know to set 'compiler' and the script errored at line 31.

climbfuji: I see - do you want me to make this change?

GeorgeGayno-NOAA: > I see - do you want me to make this change?

Please.  That would help the user.  The default does not have to be intel.  You could also do something like "compiler=${compiler:?}".   Anything to alert the user to set 'compiler'.

JeffBeck-NOAA (approved):

build_all.sh
JeffBeck-NOAA: @DomHeinzeller, @GeorgeGayno-NOAA, I just merged this PR, but then noticed this comment wasn't resolved.  Does this change still need to be made?  Thanks.

climbfuji: Not sure, I lost track because this has been sitting there for so long. But the PR hasn't been approved by @GeorgeGayno-NOAA or @LarissaReames-NOAA and better should not have been merged?

Issue #191, opened 2020-10-27
Add doxygen documentation


edwardhartnett:
See https://github.com/NOAA-EMC/NCEPLIBS/issues/121

edwardhartnett:
Are we all done with this issue? I think there is still more to do...

GeorgeGayno-NOAA:
> Are we all done with this issue? I think there is still more to do...

I just opened another branch to add doxygen to the remaining dozen or so files.   So keep it open.

edwardhartnett:
We need to add an overview of UFS_UTILS in the user_guide.md. This will include a (hyper-lined) list of all utilities, with a brief description.

I will submit a PR shortly. @GeorgeGayno-NOAA I'm going to need your help coming up with the descriptions...

GeorgeGayno-NOAA:
> We need to add an overview of UFS_UTILS in the user_guide.md. This will include a (hyper-lined) list of all utilities, with a brief description.
> 
> I will submit a PR shortly. @GeorgeGayno-NOAA I'm going to need your help coming up with the descriptions...

No problem.

edwardhartnett:
@lgannoaa, we need doxygen for UFS_UTILS on
* Every function/subroutine 
* Every public variable
* Every code file 

(Except for testing code, which is documented but not with doxygen. We do not include a description of our test code in the user documentation.)

More details about Doxygen can be found here:
https://github.com/NOAA-EMC/NCEPLIBS/issues/121

@jbathegit has a tip sheet which is posted in that thread. Jeff is very far along in applying doxygen to NCEPLIBS-bufr, and that's a great place to look for examples.

GeorgeVandenberghe-NOAA:
Per suggestion I have (had already) been ignoring module CONTAINED docblock rendering  issues (while addressing other issues in those docblocks) since wednesday afternoon.  So this did not stop me.

edwardhartnett:
Not sure hot this issue got closed. We will keep it open until we turn on WARN_AS_ERROR everywhere (perhaps with some exceptions), and fill in all "???" in the current documentation.

edwardhartnett:
It's been a heck of an issue!

Documentation work will continue but I believe this issue can be called complete.

PR #195, opened 2020-10-28
more doxygen work

edwardhartnett:
Part of #191 

I have built the current doxygen docs, and put them up at https://noaa-emc.github.io/UFS_UTILS/. I will update as we develop these docs. Eventually, this on-line version will only be updated when we do a version release, but while the docs are under development, I will keep it updated when there are changes.

In this PR I add the list of source directories to the Doxyfile.in. If directories are added, renamed, or deleted, the Doxyfile.in will also need to be updated so it can find the code.

In our code, we must convert existing comments to doxygen comments. I have done so for one file in this PR.

Once we get this PR merged I will convert the other Fortran files, in batches, trying to avoid any active work areas so there are no conflicts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@
 Utilities for the NCEP models. This is part of the
 [NCEPLIBS](https://github.com/NOAA-EMC/NCEPLIBS) project.
 
+Complete documentation can be found at
+https://noaa-emc.github.io/UFS_UTILS/.
+
 ## Authors
 
 NCEP/EMC developers.
--- a/docs/Doxyfile.in
+++ b/docs/Doxyfile.in
@@ -741,7 +741,23 @@ WARN_LOGFILE           =
 # spaces.
 # Note: If this tag is empty the current directory is searched.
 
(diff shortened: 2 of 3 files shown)


GeorgeGayno-NOAA:
@edwardhartnett When I look at the list of files - https://noaa-emc.github.io/UFS_UTILS/files.html -  some routines have the source directory preceding the name.  But for most routines you can't tell what program they belong to.  Can the list be sub-divided by program or is that a follow-on PR?

edwardhartnett:
That list will not be sub-divided, but we may present whatever organization we like on the mainpage: https://noaa-emc.github.io/UFS_UTILS/index.html

Right now, the mainpage says very little, but we can put what we like there, and organize the lists of functions into whatever lists and tables seem best.

But first we must convert existing comments to doxygen, and these will show up in the files list. If you look at the current list, only nst_tf_chg.f90 can be clicked, because that's the one file I converted to doxygen.

Let's merge this PR and I will convert some more files.

GeorgeGayno-NOAA (approved):
Looks good.

GeorgeGayno-NOAA:
@edwardhartnett You should probably create your own fork for these branches.

PR #198, opened 2020-10-28
More doxygen work

edwardhartnett:
Part of #191 

More work on the doxygen documentation.

### DIFF
--- a/sorc/sfc_climo_gen.fd/driver.F90
+++ b/sorc/sfc_climo_gen.fd/driver.F90
@@ -1,24 +1,22 @@
- program driver
-  
-!-------------------------------------------------------------------------
-!  program documentation block
-!
-! Abstract: Reads static surface data on a global lat/lon grid,
-!   interpolates the data to the fv3 model grid, and outputs the
-!   result in netcdf format.
-!
-! Program execution is controlled by variables defined in the
-! program configuration namelist (see module program_setup for
-! details).
-!
-! Requires the following input files:
-!   1) Model mosaic file (netcdf)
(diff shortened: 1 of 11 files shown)


GeorgeGayno-NOAA (approved):
Looks ok.

edwardhartnett:
Are you looking at the output as well? You should do so, it is here: https://noaa-emc.github.io/UFS_UTILS/.

GeorgeGayno-NOAA:
> Are you looking at the output as well? You should do so, it is here: https://noaa-emc.github.io/UFS_UTILS/.

I see that more of the files are now 'clickable'.  And the info in most of them looks ok.   However, I tend to recycle module names.  You will see multiple instances of  'driver.f90' and 'program_setup'.  When I click on https://noaa-emc.github.io/UFS_UTILS/sfc__climo__gen_8fd_2program__setup_8f90.html  Then click on the module program setup link (https://noaa-emc.github.io/UFS_UTILS/namespaceprogram__setup.html), it is pointing to the program-setup under chgres_cube, not sfc_climo_gen.  So I think it is getting confused by the duplicate files names.  I hope we don't have to rename all the modules with unique names.

edwardhartnett:
We will not have to rename modules. Yes, there is some confusion about this in the output, but I'm sure we can straighten that out. I will do some browsing of on-line resources and see if I can figure it out.

But the files list is actually not the way most readers will interact with the docs. We will add a mainpage  that will be more user-friendly face on the docs.

I will merge this PR and add something to the mainpage for the next round, so you can see what I mean...

Issue #207, opened 2020-11-02
Only build the required UFS_SRW utilities in release/public-v2


mkavulich:
The release/public-v2 branch is specific to the UFS Short Range Weather App release (https://github.com/ufs-community/ufs-srweather-app) and so it should not build unnecessary utilities.

mkavulich:
Yes, thanks for the reminder @GeorgeGayno-NOAA. We now only build the necessary utilities in the release branch, per PR #208

PR #208, opened 2020-11-03
Do not build utilities that are not needed for the UFS SRW App

mkavulich:
This PR is simple: I just comment out the following utilites from the CMakeLists.txt files that are not needed for the UFS SRW App:
```
  emcsfc_ice_blend
  emcsfc_snow2mdl
  fregrid
  fvcom_to_FV3
  global_chgres
  global_cycle
  inland
  lakefrac
  mkgfsnemsioctl
  nemsio_chgdate
  nemsio_get
  nemsio_read
  nst_tf_chg
```
Tests run: built UFS_UTLS on Macos (gnu 10.2) and Hera (intel/18.0.5.274); the list of executables is now as expected, only the necessary 10 utilities are built:
```
  chgres_cube
  filter_topo
  global_equiv_resol
  make_hgrid 
  make_solo_mosaic
  orog
  regional_esg_grid
  sfc_climo_gen
  shave
  vcoord_gen
```
Also ran end-to-end build/run tests of the UFS SRW App on Hera, all completed successfully.

Resolves issue #207


GeorgeGayno-NOAA:
@mkavulich Be aware that the 'inland' and 'lakefrac' programs are run (optionally) as part of the grid generation scripts (see ./driver_scripts).  In the release branch, the lake option is only supported for global uniform grids and not regional grids.  (The default is no lakes.)  So as long as users only create regional grids, this is not a problem.

mkavulich:
@GeorgeGayno-NOAA Sorry for the delay in responding. This is specific to the SRW release branch, so it should not be used by users trying to run on global uniform grids. Additionally, we do not plan on documenting these capabilities for the release, so more advanced users who may want to delve into these options for regional grids should use `develop` anyway.

GeorgeGayno-NOAA (approved):
I checked out his branch on WCOSS-Dell and ran the build script.  Only programs needed for UFS SRW App were built as expected.  Will merge.

PR #221, opened 2020-11-12
Changes from nwprod to nwpara for wgrib2 v2.0.8 library

MatthewPyle-NOAA:
Is a temporary change, but right now chgres_cube won't build with the existing module file.  I'm planning to deliver to NCO early next week, so would like to give them something that will build out of the box.


GeorgeGayno-NOAA (approved):
Matt, no problem.  We can adjust it again when wgrib2 is finally installed.  (Hopefully, soon).

Issue #222, opened 2020-11-17
chgres_cube : Bug in reading grib2 vertical coordinate type


LarissaReames-NOAA:
Operational RAP grib2 files on the native grid (and possibly other grib2 data types) are provided on isobaric coordinates, but contain some data on the 1.0 hybrid coordinate. This results in the code proceeding as if all of the data were on hybrid coordinates. Should be remedied with a change to the "hybrid" search string. 

This issue should be corrected in both develop and release/public-v2 branches.

LarissaReames-NOAA:
Merged in both develop and public/release-v2

PR #223, opened 2020-11-17
wgrib2 text search fix to account for isobaric files 

LarissaReames:
Some isobaric wgrib2 files have data on some hybrid levels. This addresses waves on boundaries when using operational RAP data for LBCs. All current and preliminary future develop branch regression tests passed on Hera.

Addresses Issue #222


GeorgeGayno-NOAA (approved):
I compiled the branch at 88717ff on WCOSS-Dell.  All chgres_cube regression tests passed.

Issue #152, opened 2020-09-22
update modulefiles to use hpc-stack


aerorahul:
The UFS-utils repository loads modules from the soon-to-be-deprecated modules installation maintained by NCEPLIBS team.
These modulefiles need to be updated to use the hpc-stack modulefiles.

The UFS-weather-model has moved to the hpc-stack and can be used as a guide for this work.

Info on the stack: https://github.com/NOAA-EMC/hpc-stack/wiki/Official-Installations

GeorgeGayno-NOAA:
Will the hpc-stack contain the wgrib2 library?  I checked on Dell (/usrx/local/nceplibs/dev/hpc-stack/libs/hpc-stack/v1.0.0-beta1/ips-18.0.1.163) and did not see it.  Also, folks will also want the wgrib2 executable.  That was not under ./grib_util/bin.  I did not check the other machines.

kgerheiser:
wgrib2 library and executable are included

GeorgeGayno-NOAA:
> wgrib2 library and executable are included

I see that now.  I was not loading the module files correctly.  Thanks.

GeorgeGayno-NOAA:
I was able to get it to build on Hera.  Yeah!

GeorgeGayno-NOAA:
Note to self: the 'nccmp' utility is part of the hpc stack.  That utility is used by the regression tests.  On some machines, I keep my own copy of it.  Use the hpc stack version.

GeorgeGayno-NOAA:
My branch at 4613fc2 builds on Hera, Dell and Orion.  All regression tests passed.

GeorgeGayno-NOAA:
@JeffBeck-NOAA @climbfuji @LarissaReames-NOAA - Under this issue, I made changes to the build script.  I found that the logic setting the CMake compiler variables is not needed on the five machines I officially support.  But I don't know if this will break the other machines, such as odin, stampede or cheyenne.  Who has access to these machines?  And can they do a quick test of my branch?  Setting these CMake flags in the build script is probably not a good practice.  Instead, if needed, they could be set in the build modules.

 ```
CMAKE_FLAGS="-DCMAKE_INSTALL_PREFIX=../ -DEMC_EXEC_DIR=ON"

-if [[ "$compiler" == "intel" ]]; then
-  if [[ "$target" != "wcoss_cray" && \
-        "$target" != "gaea" && \
-        "$target" != "odin" ]]; then
[...]

PR #227, opened 2020-11-18
Use hpc-stack in UFS_UTILS

GeorgeGayno-NOAA:
The Hera build already uses the stack.  Add a stack build for Jet, Orion 
and WCOSS-Dell.  The stack is unavailable on WCOSS-Cray.

For details, see #152


GeorgeGayno-NOAA:
@kgerheiser You mentioned a new release of hpc-stack soon.  Should I hold off on this merge?  Currently, UFS_UTILS does not work on Jet and Orion.  So updating to the stack is important.

kgerheiser:
It's up to you. We'll have that installed in a day or two. You could go ahead with it and then we'll let you know when the update is ready.

Now that you have it building with hpc-stack updating to a newer version should be as easy as changing a few lines in the modulefiles.

kgerheiser (approved):

Issue #228, opened 2020-11-20
chgres_cube: Various small bug fixes and sanity check additions


LarissaReames-NOAA:
A few small issues to fix, none of which seemed to deserve its own issue:

1) Various extraneous print statements that are left over from debugging should be removed.

2) There's a bug when using HRRR or RAP grib2 input and climatological surface fields (*_from_climo=true) where the incorrect data is saved to memory as the soil type.

3) If an incorrectly-formatted or empty variable mapping table (for grib2 data) is provided, the code does not fail. This needs to be changed, with a failure at read with a descriptive message.

4) When reading soil type from geogrid files for HRRR or RAP grib2 input data, there is no check for grid shape, so the user does not know if they've provided the incorrect geogrid file. Add a check on grid size to account for this. Also, adjust logic in this section to cause execution failure if any netcdf call returns an error. 

All of these issues will be addressed in a single PR.

PR #229, opened 2020-11-20
release/public-v2: Various chgres_cube bug fixes and code improvements

LarissaReames-NOAA:
This PR addresses all of the listed items in Issue #228. Code compiles and passes all current and future regression tests on Jet.

### DIFF
--- a/sorc/chgres_cube.fd/atmosphere.F90
+++ b/sorc/chgres_cube.fd/atmosphere.F90
@@ -1163,9 +1163,6 @@
  endif
 
  print*
- do k = 1, levp1_target
-    print*,'VCOORD FOR LEV ', k, 'IS: ', vcoord_target(k,:)
- enddo
  
  close(14)
 
--- a/sorc/chgres_cube.fd/atmosphere.F90
+++ b/sorc/chgres_cube.fd/atmosphere.F90
@@ -1162,8 +1162,6 @@
    call error_handler("READING VERTICAL COORD FILE", istat)
  endif
 
- print*
- 
  close(14)
 
  end subroutine read_vcoord_info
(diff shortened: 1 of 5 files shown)


sorc/chgres_cube.fd/atmosphere.F90
```diff
@@ -1163,9 +1163,6 @@ subroutine read_vcoord_info
  endif
 
  print*
```
JeffBeck-NOAA, line 1165: Remove line 1165 too?

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -4956,6 +4958,8 @@ subroutine read_input_sfc_grib2_file(localpet)
      where(slmsk_save == 1) dummy2d_i = 1
    
      call search(dummy2d_8,dummy2d_i,i_input,j_input,1,230)
+   else
```
JeffBeck-NOAA, line 4961: What's this do?

sorc/chgres_cube.fd/atmosphere.F90
LarissaReames-NOAA: Oh yep, good catch. I'll take that out.

sorc/chgres_cube.fd/input_data.F90
LarissaReames-NOAA: This fixes a problem I encountered using HRRR data with surface climo fields. The input data from wgrib2 is 4-bit, but esmf expects 8 bit, so all data has to be converted using the real() command, saving the 8-bit form of the data from dummy2d in to dummy2d_8, which is then broadcast using esmf. Before I added this, if sotyp_from_climo = true, the dummy2d_8 array wouldn't be re-populated with the correct soil (or fill) data from dummy2d, and it would still have the data from the previous use of dummy2d_8, which was skin temperature.

JeffBeck-NOAA: Thanks! Good catch!

JeffBeck-NOAA (approved):

Issue #235, opened 2020-12-10
Remove module.base files


GeorgeGayno-NOAA:
As far as I know, these files are not used. (@KateFriedman-NOAA verified they are not used by global workflow).  Get rid of them.

PR #237, opened 2020-12-15
Remove unused module files

GeorgeGayno-NOAA:
Remove unused module files, and an obsolete global_chgres script that used these module files.

Issue #235


KateFriedman-NOAA (approved):
Cloned branch on Orion and ran build_all.sh. Built without issue. I notice that it is using v1.0.0-beta1 hpc-stack, very nice! I approve this cleanup, thanks George!

Issue #243, opened 2021-01-21
orog program fails on MacOS with gfortran, `Illegal instruction: 4`


mkavulich:
I have been trying to debug this problem for a while but have had no success. Running orog on MacOS Catalina (10.15.7) compiled with gfortran 9.3.0 (GNU Fortran (MacPorts gcc9 9.3.0_4) 9.3.0) fails with the message `Illegal instruction: 4`. This issue occurs with or without DEBUG settings turned on (there seems to be no difference in behavior regardless of compilation flags). 

This is the end of the output leading up to failure:

```
  Before GICE ZAVG(1,2)=        2801           1
  Before GICE ZAVG(1,12)=        2804           1
  Before GICE ZAVG(1,52)=        2789           1
  Before GICE ZAVG(1,112)=           0           1
  GICE 30" Antarctica RAMP orog 43200x3616 read OK
  Processing! 
  Processing! 
  Processing! 
  Ocean lsm file Opened OK: mskocn,istat=           1           0
  Rd fail: Ocean lsm - continue, mskocn,ios=           0          -1
 outgrid=/Volumes/d1/workdir/UFS/one_more_macos_try/ufs-srweather-app/experiment_DEBUG_FLAGS/test_GSD_HRRR_AK_50km/fix_lam/C201_grid.tile7.halo6.nc
 Read the grid from file /Volumes/d1/workdir/UFS/one_more_macos_try/ufs-srweather-app/experiment_DEBUG_FLAGS/test_GSD_HRRR_AK_50km/fix_lam/C201_grid.tile7.halo6.nc
 minlat=   44.683594338329847      maxlat=   76.964315294499997     
 north pole supergrid index is            0           0
 south pole supergrid index is            0           0
  Timer 1 time=    55.631000000001222     
Illegal instruction: 4
```

Based on the output, the failure appears to be in the calling or initialization of the MAKEMT2 subroutine, but somehow I can not tease any more debugging info: the executable does not print a traceback or any other information aside from "Illegal instruction: 4".

[Here is the full log](https://github.com/NOAA-EMC/UFS_UTILS/files/5852235/orog.log), everything looks normal until the sudden failure.

I'm at a loss as to what to try next, so I'm hoping someone else can offer some suggestions, or better yet try this themselves and try to reproduce the issue. Let me know if you need any more info.

climbfuji:
[...]

PR #245, opened 2021-01-22
Rearrange allocation of hgt_1d in mtnlm7_oclsm.f to fix errors on MacOS

mkavulich:
There are some declarations of allocatable variables that result in `illegal instruction` errors on MacOS, despite being valid Fortran77 code. One such instance is found in `sorc/orog_mask_tools.fd/orog.fd/mtnlm7_oclsm.f`, where the real `hgt_1d` is declared in the subroutine MAKEMT2. While the existing code is valid, it can be rearranged in a way that does not trigger this error but gives equivalent results.

This code has been tested on MacOS (GNU 9.3.0), Cheyenne (GNU 10.1.0) and Hera (Intel 18.0.5) by running the end-to-end UFS SRW App workflow release branch tests. Also checked on Cheyenne to ensure the output files from `orog` for the GST test case were bit-for-bit identical. Let me know if there are any other tests required for this PR to be accepted.

This PR will resolve Issue #243. Thanks to @climbfuji for providing the breadcrumbs needed to find and fix this problem.

### DIFF
--- a/sorc/orog_mask_tools.fd/orog.fd/mtnlm7_oclsm.f
+++ b/sorc/orog_mask_tools.fd/orog.fd/mtnlm7_oclsm.f
@@ -1902,9 +1902,9 @@ C  (*j*)  for hard wired zero offset (lambda s =0) for terr05
          ENDDO
       ENDDO
 !$omp end parallel do
-      WRITE(6,*) "! MAKEMT ORO SLM VAR VAR4 DONE"
+      WRITE(6,*) "! MAKEMT2 ORO SLM VAR VAR4 DONE"
 C
-
+      deallocate(hgt_1d)
       RETURN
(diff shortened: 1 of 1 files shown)


climbfuji (commented):
I would explicitly add the deallocate statement at the end of this routine.

GeorgeGayno-NOAA:
This fix should added 'develop' as well?

mkavulich:
@GeorgeGayno-NOAA Yes I believe this should be added to develop as well, though I have not specifically tried develop on MacOS yet.

mkavulich:
@climbfuji I added the deallocate statement, and re-tested on MacOS and Hera, all successful.

climbfuji (approved):

Issue #246, opened 2021-01-25
chgres_cube, release/public-v2: Soil type issue at land ice vegetation points


LarissaReames-NOAA:
Due to inconsistencies in landmask between HRRR and FV3, and how HRRR data handles vegetation and soil type at frozen lakes / land ice points, some points exist in surface output data with soil type of 0 at land ice points. 

The fix for this, while not ideal, is a brute force method of forcing all grid points with land use type of land ice to have soil type 16 in the qc_check subroutine. I have been unsuccessful, despite many many efforts, to implement a fix by modifying the input data just after reading in, or by making changes in the interp subroutine. Will use this brute force fix for now, but will contact ESMF support before implementing a fix in the develop branch.

GeorgeGayno-NOAA:
Do you know why the problem (soil type 0) only happens at land ice points?

lharris4:
Hello. I don't understand how this is a problem with FV3. FV3 is a
dynamical core and has no conception of vegetation type, lakes, or land ice.

LarissaReames-NOAA:
> Do you know why the problem (soil type 0) only happens at land ice points?

It's honestly a complete mystery. I've tracked the total number of problem points on the target grid (by checking places where soil type is 0 over land points) and they are remedied by doing a soil type interpolation along with the other land ice  to land ice interpolation calls, but then they reappear (the exact same number) after the land, non-land ice interpolation. I confirmed that there are no soil type of 0 at any land points on the input grid, so I'm honestly not sure where those points are originating from.

LarissaReames-NOAA:
> Hello. I don't understand how this is a problem with FV3. FV3 is a dynamical core and has no conception of vegetation type, lakes, or land ice.
> […](#)

Larry, this is the UFS_UTILS GitHub, not FV3.

lharris4:
I am referring to this line in the original message.

> Due to inconsistencies in landmask between HRRR and FV3, and how HRRR
data handles vegetation and soil type at frozen lakes / land ice points,
some points exist in surface output data with soil type of 0 at land ice
points.

LarissaReames-NOAA:
> I am referring to this line in the original message.
> Due to inconsistencies in landmask between HRRR and FV3, and how HRRR
> data handles vegetation and soil type at frozen lakes / land ice points, some points exist in surface output data with soil type of 0 at land ice points.
> […](#)

[...]

PR #247, opened 2021-01-26
Soil type at land ice fix when using soil and vegetation type from input data (HRRR/RAP)

LarissaReames-NOAA:
chgres_cube fix for soil type == 0 at land ice points when using soil and vegetation type from input data.

Addresses Issue #246.


JeffBeck-NOAA:
Tested successfully on Hera.  Approving and merging now.

JeffBeck-NOAA (approved):

Issue #256, opened 2021-02-08
write fortran unit test for chgres_cube.fd/input_data.F90


edwardhartnett:
We need to start writing fortran unit tests. THis is as good a place to start as any...

edwardhartnett:
As discussed in the recent meeting, George will be testing this next as part of helping with testing with @LarissaReames-NOAA changes.

GeorgeGayno-NOAA:
Next, I will write a test for routine "convert_winds".

GeorgeGayno-NOAA:
Should I create a unit test for one of the read routines next? `read_input_nst_netcdf_file` is one of the shorter ones. There is not much to these read routines. They read records from a file, then place the data in the appropriate ESMF field. What should constitute "success"? Create 'ncdump' files of each record and compare against the corresponding ESMF field? Just check a few points against expected values?

edwardhartnett:
Checking a few points against expected values is a good start.

Success should be when you check the ESMF field and it has the value that was in the file.

Are these large input data files?

edwardhartnett:
In terms of order of testing, I really have no preference. We need to test all of it. I'm working on some of the write functions - anything else you see, have at it. Just enter an issue with the name of the program file in the title (like this issue) and then assign it to yourself. That way, we can avoid writing tests for the same code. ;-)

GeorgeGayno-NOAA:
> Checking a few points against expected values is a good start.
> 
> Success should be when you check the ESMF field and it has the value that was in the file.
> 
> Are these large input data files?

The surface NetCDF file is 1 GB. The atmospheric file, which is read by another routine in this module, is 6.9 GB. I wonder if Github will choke on this.

edwardhartnett:
You cannot check those into github and they are too big to FTP. ;-)

We could cut much of the data from the file and write a test based on the stripped-down file...

GeorgeGayno-NOAA:
> You cannot check those into github and they are too big to FTP. ;-)
> 
> We could cut much of the data from the file and write a test based on the stripped-down file...

Sorry for the confusion. I did not intend to add the file to the repository. I just wonder if a unit test that reads that much data will max out github somehow.

edwardhartnett:
GitHub maxes out at 30 minutes, so that's a lot of data reading.

But do we need to read the whole file in order to test the code? Probably not.

Also, we cannot upload a 1 GB file to GitHub for testing. It will take too long to FTP that big a file. We need to test this code with a smaller file.

[...]

PR #264, opened 2021-02-08
adding test code

edwardhartnett:
Part of #256 

Starting to add unit test for some fortran code.


edwardhartnett:
@lgannoaa this is an example of how to add an F90 test to the code. Right now my test does nothing. (I'm working on the caching issue in the CI to make this go faster).

edwardhartnett:
I am trying to get ESMF to build on my local machine. When I do I can finally get this test built on my local machine and start adding code to it...

edwardhartnett:
That is awesome. @GeorgeGayno-NOAA you will be interested in this. 

Currently lots of our docblocks laboriously list functions that use or are used by the documented function. Such lists are notoriously inaccurate in practice, and a pain to maintain. Now we can generate them, with no maintenance effort, and they will always be correct and complete!

Very nice!

edwardhartnett:
Opps, the comment above was meant to be in the NCEPLIBS-bufr repo. ;-)

I will close this PR because Lin is about to put one up with a working test...

PR #272, opened 2021-02-09
updated workflow to enable caching of ESMF

### LINKED ISSUE #260: Modify existing workflow to use caches for ESMF, NCEPLIBS, etc.
Currently we are rebuilding the same versions of ESMF, Jasper, and NCEPLIBS.

Seems like we could cache these, since they are the same every time. This would make the CI faster.

edwardhartnett:
Fifth time's the charm!

Part of #260 

It took me many tries to get the right. Turns out that GitHub actions sometimes just freezes when your workflow file is wrong. ;-)

In this PR I change the CI to get ESMF and other codes with wget, instead of checking them out, and then I use GitHub caching, so that these codes are only build the first time. For subsequent tests, GitHub pulls them from the cache.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -22,7 +22,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ~/esmf
-        key: esmf-${{ runner.os }}-8.0.1
+        key: esmf-${{ runner.os }}-8.0.1-1
 
     - name: build-esmf
       if: steps.cache-esmf.outputs.cache-hit != 'true'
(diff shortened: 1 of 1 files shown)


edwardhartnett:
OK, this is ready to merge and is an improvement. @kgerheiser you may want to take a look to see if you think I messed anything up here. Basically, I moved where ESMF was installed, cached it, and now use a wget to get a tarball instead of checking out a tagged version of the ESMF code.

The end result is that ESMF will no longer be built each time the test is run.

After this PR is merged I will take another stab and doing this for the jasper and nceplibs in this CI as well. But after all this dorking around, I want to at least get the ESMF part into develop branch! ;-)

.github/workflows/build_and_test.yml
```diff
@@ -77,7 +84,8 @@ jobs:
 
     - name: build-ufs-utils
       run: |
-        export ESMFMKFILE=~/lib/esmf.mk
+        export ESMFMKFILE=~/esmf/lib/esmf.mk
+        export ESMF_DIR=~/esmf-ESMF_8_0_1
```
kgerheiser, line 88: Why do you need to set ESMF_DIR here?

edwardhartnett: Maybe I don;t. Let me test it...

edwardhartnett: I don't!

kgerheiser (approved):

GeorgeGayno-NOAA:
I see you are now saving 10 minutes of compile time.  Nice.

GeorgeGayno-NOAA (approved):

PR #274, opened 2021-02-10
for GitHub workflow,s, use caching, rename for clarity, add workflow for nceplibs develop branch test

### LINKED ISSUE #259: New workflow to build against develop branch of NCEPLIBS
We are testing against v1.2.0 of NCEPLIBS.

This is good and necessary.

We should also be testing against the development branches of all the NCEPLIBS. That way, if one of them changes in a way that causes a UFS_UTILS test to fail, we find out immediately, instead of after the next time we do a release. (Just exactly too late.) ;-)

@lgannoaa if you want to take this on it will start by adding a new file in .github/workflows. You will have to read up on github workflows, because this is a technology integral to our testing efforts.

### LINKED ISSUE #260: Modify existing workflow to use caches for ESMF, NCEPLIBS, etc.

edwardhartnett:
Fixes #260 
Fixes #259

Now cache jasper build.
Also rename workflows for clarity.
Also added workflow to test against nceplibs-1.3.0.
Changed existing workflow to use nceplibs develop branch.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -54,6 +54,9 @@ jobs:
 
     - name: build-jasper
       run: |
+        set -x
+	pwd
+	wget https://github.com/jasper-software/jasper/archive/version-2.0.22.tar.gz &> /dev/null && ls -l
         cd jasper
         mkdir build-jasper && cd build-jasper
         cmake .. -DCMAKE_INSTALL_PREFIX=~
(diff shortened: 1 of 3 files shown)


edwardhartnett:
@GeorgeGayno-NOAA if we can get this reviewed and merged this morning, it will be helpful in test development...

kgerheiser (approved):

GeorgeGayno-NOAA (approved):

GeorgeGayno-NOAA:
@edwardhartnett  When I check the history of 'develop' I see all your branch commits.  Can you do a squash merge next time?

PR #276, opened 2021-02-10
Feature/unit tests for chgres_cube.fd/utils.f90

### LINKED ISSUE #257: Write fortran tests for chgres_cube.fd/utils.f90
Write a simple program to test to_upper() and to_lower() functions.

lgannoaa:
Thanks for a great team work help from @GeorgeGayno-NOAA.
This commit is the draft for unit test on to_upper and to_lower function from chgres_cube.fd/utils.f90.
Running this application "to_upper_lower" under exec directory. Result will print as Pass to indicate a good testing.

Fixes #257


unit-tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+set(fortran_src
```
GeorgeGayno-NOAA, line 1: Is there a better way to point to source code in another directory?

unit-tests/to_upper_lower.F90
```diff
@@ -0,0 +1,48 @@
+!> @file
```
edwardhartnett, line 1: Please rename this directory from "unit_tests" to simply "tests", which is the cmake convention.

unit-tests/to_upper_lower.F90
```diff
@@ -0,0 +1,48 @@
...
+
+!> output
```
edwardhartnett, line 3: Actually convert these to non-doxygen comments. All production code gets doxygen, but not test code. ;-)

ALso add your name as the author of this test.

unit-tests/to_upper_lower.F90
```diff
@@ -0,0 +1,48 @@
...
+ print*, "------------ Testing to_lower and to_upper --------------"
+
+!-------------------------------------------------------------------------
+! Execute testing below by running target function with testing string 
+!   When match_result set to be T - compare to base line is identical
+!   When match_result set to be F - compare to base line is NOT identical
+!-------------------------------------------------------------------------
+
+ call to_lower(test_input_char_1) 
+ match_result = test_input_char_1 == l_st_base
```
edwardhartnett, line 33: If this is not correct, fail immediately. In general we want to stop a test as soon as we have found a problem. You can just use a stop here...

edwardhartnett (commented):
If you merge in the develop branch you will get the new GitHub workflow files. Add "make test" to the bottom of the file and run your test in the CI.

unit-tests/CMakeLists.txt
lgannoaa: I found a solution by change this line from:
../sorc/chgres_cube.fd/utils.f90
to
${CMAKE_SOURCE_DIR}/sorc/chgres_cube.fd/utils.f90
I tested and it works. Is this solution to your satisfaction?

edwardhartnett (approved):

edwardhartnett:
OK, I made a few more changes to this branch:
* Renamed the test file to ftst_utils.f90. (Let's use convention that fortran tests start with "ftst_" and C tests start wtih "tst_").
* Changes the CMake file to build ftst_utils as a test, not an installed executable, and to run it as a test.
* Changed workflows to now run tests (which they could not do before, since there were no tests!)
* Change test code documentation to not be doxygen (we won't use doxygen on test code, but it should still have good comments).
* Changed output. Test output should like like this for us:

```
Testing some group of features.
testing feature 1... 
ok
testing feature 2...
ok
SUCCESS!
```

This is what I use in netcdf-c, netcdf-fortran, PIO, and other libraries. It works well and looks nice. It's reassuring without being too much to read. ;-)

In fact, the test output is only seen when the user manually runs a test. In the normal case, make test is used, and cmake runs all the tests, and the user does not see any test output, only whether it passed or failed.

edwardhartnett:
One last change, I moved the test to a chgres.cube.fd directory under the tests directory. We can have a sub-directory in there for each utility, just like under sorc.

edwardhartnett:
OK, I promise, no more tinkering with this one. ;-)

@GeorgeGayno-NOAA and @lgannoaa please take a look at how this ended up. If you two are happy with it, let's get this PR merged and move on to more testing!

Great first test from @lgannoaa to get us started off. I've added testing to many projects, and the first test is always the hardest!

edwardhartnett:
Also I just remembered that George asked us to do squashed commits. That is, commits that will show up in his history as just one change, instead of keeping a record of all our incremental changes.

I'm down with that, but I don't know how to do it. I thought there would be a button here on the pull request, but I can't find it...

GeorgeGayno-NOAA:
> Also I just remembered that George asked us to do squashed commits. That is, commits that will show up in his history as just one change, instead of keeping a record of all our incremental changes.
> 
> I'm down with that, but I don't know how to do it. I thought there would be a button here on the pull request, but I can't find it...
> 
> OK, found it, in the merge button. So George, I think this is something you do when you merge this PR...

Whenever I do a merge, the squash commit is the always default merge button.  I would think it would be the default for everyone.

kgerheiser:
It defaults to the last one you used.

edwardhartnett:
I will go ahead and merge this, since I think we're all good with it... (I will squash the commits.)

Issue #167, opened 2020-10-13
Remove nemsiogfs library from global_chgres and all other places


edwardhartnett:
The NCEPLIBS-nemsiogfs library is scheduled for deprecation with the next release of NCEPLIBS. We believe this will not be a problem for any users, but it needs to be removed wherever it is offered as an option in UFS_UTILS.

See https://github.com/NOAA-EMC/NCEPLIBS/issues/126 for detailed discussion.

Please speak up now if you think this will be a problem for anyone. ;-)

@Hang-Lei-NOAA you will be interested in this issue.

Hang-Lei-NOAA:
Thanks, George. So we can deprecate it without any problem.

edwardhartnett:
Whenis global_chgres going to stop being used?

GeorgeGayno-NOAA:
I am not aware that anyone is using it.  And the best way to prevent future users is to get rid of it now.

edwardhartnett:
OK, shall I submit a PR removing it's directory and all contents?

GeorgeGayno-NOAA:
> OK, shall I submit a PR removing it's directory and all contents?

If you want.  There are several associated scripts that should be removed as well.

edwardhartnett:
Which scripts?

GeorgeGayno-NOAA:
> Which scripts?

./ush/global_chgres.sh
./ush/global_chgres_driver.sh

PR #299, opened 2021-02-17
delete global_chgres

### LINKED ISSUE #298: remove global_chgres and associated scripts
As discussed in #167.

Remove global_chgres and scripts:
./ush/global_chgres.sh
./ush/global_chgres_driver.sh

edwardhartnett:
Fixes #298 
Part of #167 

Removed global_chgres.


edwardhartnett:
You know maybe we should not be doing this at all.

If we are still supporting this in public-v2, then we are just going to have to maintain it on that branch. We might as well have it on develop as well, for easy testing...

edwardhartnett:
@GeorgeGayno-NOAA OK, if we're going to do this, I guess we should do this.

Once we delete global_chgres it will only be maintained on the public-v2 branch. It will receive no help with documentation and testing.

edwardhartnett:
@GeorgeGayno-NOAA shall we merge this? You must review it...

GeorgeGayno-NOAA (changes requested):
global_chgres is the only ufs_utils program that uses the 'landsfcutil' and the 'nemsiogfs' libraries.  That what originally prompted this task - to get rid of nemsiogfs.  So you can eliminate the find_package calls for those libraries.  And remove their module load from all files in ./modulefiles.

edwardhartnett:
@GeorgeGayno-NOAA I've removed nemsiogfs and landsfcutil from the CMake and module files...

GeorgeGayno-NOAA:
Let me do a quick compile on all our machines.

GeorgeGayno-NOAA (approved):
Successfully built with Intel on WCOSS-Cray, WCOSS-Dell, Jet and Orion.   Successfully built with Gnu and Intel on Hera.

Issue #218, opened 2020-11-12
Recover test from sorc/fre-nctools.fd/shared_lib/create_xgrid.c


edwardhartnett:
There is some test code in sorc/fre-nctools.fd/shared_lib/create_xgrid.c that looks like it should be a unit test.

PR #301, opened 2021-02-17
trying to build test for fre-nctools.fd/shared_lib/create_xgrid.c

edwardhartnett:
Part of #218


edwardhartnett:
This needs more work in future iterations, because right now it just runs one of the test cases, and prints output, so it's not really checking the results.

However, the checking will be easier to add once we get the documentation complete. FOr now, let's merge this and work on improving it in future PRs.

Hang-Lei-NOAA (approved):

edwardhartnett:
@GeorgeGayno-NOAA were you running this test as part of your regression tests? If so, this PR will break your tests, because the test code has moved to it's own code file...

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA were you running this test as part of your regression tests? If so, this PR will break your tests, because the test code has moved to it's own code file...

My regression tests are under ./reg_tests.  Should I test them?

PR #347, opened 2021-02-23
now pass filename to read_setup_namelist()

### LINKED ISSUE #340: add filename parameter to read_setup_namelist() in chgres_cube.fd/program_setup.F90
We have function read_setup_namelist() which reads the setup namelist.

Currently this opens a file "./fort.41" to read the configuration namelist.

This is a little bit of a testing challenge. We have different namelist files we are going to need to test with. We can't name them all the same thing. We could use test shell scripts to move different namelist files around, naming whatever file we wanted for the next test "./fort.41" but this would add a lot of complexity to testing.

I propose to add an optional filename parameter to the subroutine, with will be set to "./fort.41" if the parameter is not present. The parameter will allow us to more easily test different namelist files in our tests.

This will have no effect on program operation. @GeorgeGayno-NOAA what do you think? Any objections?

edwardhartnett:
Fixes #340 

now pass filename to read_setup_namelist()


GeorgeGayno-NOAA (approved):
Ok.  I see what you are doing.  That is fine with me.

PR #359, opened 2021-02-25
Doxygen updates to chgres_cube.fd/surface.F90

### LINKED ISSUE #281: improve doxygen documentation for chgres_cube.fd/surface.F90
The first part of the testing effort is always going to be to improve the doxygen documentation. For example, in chgres_cube.fd/surface.F90 we have:

```
!---------------------------------------------------------------------------------------------
! Horizontally interpolate surface fields using esmf routines.
!---------------------------------------------------------------------------------------------

 subroutine interp(localpet)

```

Firstly, the existing documentation needs to be converted to doxygen. Then we need a param tag, an author tag (if known). More explanation would be welcome.

Consider how @lgannoaa will approach this task and put some documentation in for each subroutine or function, so that a tester will have some idea of how to start writing a test for that subroutine or function. Of course, you cannot put everything you know into documentation. But more than one sentence would be welcome.

If we can ask you @GeorgeGayno-NOAA to target each code file in chgres_cube.fd in turn, and do a doxygen upgrade, that would speed testing quite a bit, and also bring up the quality of the documentation.

GeorgeGayno-NOAA:
Minor doxygen updates.

Fixes #281

### DIFF
--- a/sorc/chgres_cube.fd/surface.F90
+++ b/sorc/chgres_cube.fd/surface.F90
@@ -20,8 +20,11 @@
 
  private
 
-! noah land ice option is applied at these vegetation types.
  integer, parameter                 :: veg_type_landice_target = 15
+                                       !< Vegetation type category that
+                                       !< defines permanent land ice points.
+                                       !< The Noah LSM land ice physics
+                                       !< are applied at these points.
 
 ! surface fields (not including nst)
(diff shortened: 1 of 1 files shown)


GeorgeGayno-NOAA:
Is there more to #281 than I realize?  I only saw minor issues.

edwardhartnett (approved):

Issue #357, opened 2021-02-25
Doxygen handling of modules not great


edwardhartnett:
When I open the doxygen output and go to files, I can navigate around and things are looking good.

But when I go to modules it does not work well. ;-)

I know there is some extra documentation glue needed here, and I will hit it after we complete our pass through UFS_UTILS, and make sure every file, module, function, and variable is documented properly.

edwardhartnett:
OK, the problem is we need two documentation blocks in each module file, one for the file, and one for the module. I will fix...

edwardhartnett:
OK, that improves the situation, and we need to do that for every module.

But that does not solve the whole problem, which stems from the fact that we are using 1 documentation build for all utilities, whereas we should be build each with its own documentation build.

If we did, then each build would have a very nice listing of modules and data types that apply only to that utility. As it is, we just throw them all in the same list, and since module names are repeated by different utilities, some get lost, and it's a mess.

PR #363, opened 2021-02-25
improving module documentation in chgres_cube

edwardhartnett:
Part of #191 
Part of #357 

Fixing doxygen handling of modules in chgres_cube.


GeorgeGayno-NOAA (approved):
With your fix, I see the 'private' routines are now displayed.  Looks good.

PR #367, opened 2021-02-26
Remove duplicate 'to_upper' from chgres_cube

### LINKED ISSUE #303: in chgres_cube, duplicate to_upper() function
In utils.f90 we have:

```
!> @brief Convert from lower to uppercase.
!! @author Clive Page
!!
!! Adapted from http://www.star.le.ac.uk/~cgp/fortran.html (25 May 2012)
!!
!! @param[in,out] strIn   string to convert
 subroutine to_upper(strIn)

     implicit none

     character(len=*), intent(inout) :: strIn
     character(len=len(strIn)) :: strOut
     integer :: i,j

     do i = 1, len(strIn)
          j = iachar(strIn(i:i))
          if (j>= iachar("a") .and. j<=iachar("z") ) then
               strOut(i:i) = achar(iachar(strIn(i:i))-32)
          else
               strOut(i:i) = strIn(i:i)
          end if
     end do
  strIn(:) = strOut(:)
end subroutine to_upper

```

In grib2_util.F90 we have:

```
!> To upper.
!! @param strIn
!! Adapted from http://www.star.le.ac.uk/~cgp/fortran.html (25 May 2012)
!! Original author: Clive Page
!! @author George Gayno  
function to_upper(strIn) result(strOut)

     implicit none

     character(len=*), intent(in) :: strIn
     character(len=len(strIn)) :: strOut
     integer :: i,j

     do i = 1, len(strIn)
          j = iachar(strIn(i:i))
          if (j>= iachar("a") .and. j<=iachar("z") ) then
               strOut(i:i) = achar(iachar(strIn(i:i))-32)
          else
               strOut(i:i) = strIn(i:i)
          end if
     end do

end function to_upper
```

GeorgeGayno-NOAA:
One version was a function and the other was a subroutine.  Only the function was being used by chgres_cube. 

The unit test was using the subroutine. So the test was updated to use the function.

Fixes #303

### DIFF
--- a/sorc/chgres_cube.fd/grib2_util.F90
+++ b/sorc/chgres_cube.fd/grib2_util.F90
@@ -88,30 +88,4 @@ subroutine convert_omega(omega,p,t,q,clb,cub)
 
 end subroutine convert_omega
 
-!> Convert string from lower to uppercase.
-!! @author Clive Page
-!!
-!! Adapted from http://www.star.le.ac.uk/~cgp/fortran.html (25 May 2012)
-!!
-!! @param[in] strIn   string to convert
-!! @return strOut string in uppercase
-function to_upper(strIn) result(strOut)
-
-     implicit none
-
-     character(len=*), intent(in) :: strIn
-     character(len=len(strIn)) :: strOut
-     integer :: i,j
-
-     do i = 1, len(strIn)
-          j = iachar(strIn(i:i))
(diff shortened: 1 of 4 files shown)


edwardhartnett (approved):
Nice! Thanks for updating the doxygen to our latest standards. That's super-helpful and something we need to do whenever we touch the code.

Issue #343, opened 2021-02-23
doxygen documentation for sorc/grid_tools.fd


edwardhartnett:
Part of #191

lgannoaa:
In review of this issue, it included 4 directories and 12 f90 source codes. I think it will be helpful if @GeorgeGayno-NOAA can take a view and input high level description document before we can finish this issue with detail doxygen update.

edwardhartnett:
We need these files doxygenated. Please add doxygen formatting as best you can. Do one or two files and put up a PR, and we will start filling in the missing information.

edwardhartnett:
I am working with Jim Purser to get the rest of these question marks filled in...

edwardhartnett:
OK, in sorc/grid_tools.fd/regional_esg_grid.fd there is one file with three remaining undocumented fields, I have sent that to Jim Purser to get the answers.

The other undocumented file is filter_topo.f90. This was written by two GFDL programmers. I have sent the file to Rusty Benson and asked them to fill in the missing fields.

edwardhartnett:
@bensonr any progress getting the programmers to document their work in filter_topo.f90?

bensonr:
@edwardhartnett  - there is a section dedicated to the topography filtering algorithm in the soon to be released comprehensive, scientific description of FV3 "the dycore" documentation.  As this offline filtering algorithm comes directly from FV3 "the dycore", that will be the authoritative source.  Once it is released, proper referencing will be available.

edwardhartnett:
@bensonr scientific documentation of the algorithm is pretty different from code documentation.

Can we have the programmers document the code as well?

PR #392, opened 2021-03-05
more doxygen for sorc/grid_tools.fd/regional_esg_grid.fd files pietc.f90 and pietc_s.f90

edwardhartnett:
Part of #191 
Part of #343

More doxygen.


edwardhartnett:
I am working with R. Purser to fill in missing info. This PR gets rid of warnings and sets the file up for R. Purser to fill in the missing info.

I am also interested in seeing how many of these constants are really used. I suspect not all of them...

GeorgeGayno-NOAA (approved):
Builds without doxygen warnings.

PR #409, opened 2021-03-11
doxygen updates in sorc/grid_tools.fd - applied comments from R. Purser

edwardhartnett:
Part of #191 
Part of #343 

More documentation filled in by R. Purser.


GeorgeGayno-NOAA (approved):
The branch compiles without doxygen warnings for this code.  The html looks good.

PR #410, opened 2021-03-11
improved chgres_cube doxygen docs with info from chgres_cube.rst, added some formatting in other utility user guides

edwardhartnett:
Part of #191 

Improving the doxygen docs.

In this PR I improve the user_guide of the chgres_cube utility, and also add a little formatting to the other users guides.

### DIFF
--- a/sorc/chgres_cube.fd/docs/user_guide.md
+++ b/sorc/chgres_cube.fd/docs/user_guide.md
@@ -9,3 +9,358 @@ This is part of the <a href="../index.html">UFS_UTILS documentation</a>.
 The chgres_cube code can be found here:
 https://github.com/NOAA-EMC/UFS_UTILS/tree/develop/sorc/chgres_cube.fd.
 
+# Introduction
+
+The chgres_cube program creates initial condition files to “coldstart”
+the forecast model. The initial conditions are created from either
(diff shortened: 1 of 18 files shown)


sorc/chgres_cube.fd/docs/user_guide.md
```diff
@@ -1,11 +1,383 @@
...
+Environmental Modeling System Input/Output (NEMSIO) data, or Network
+Common Data Form (NetCDF) data.
+
+This is part of the [NCEPLIBS
+UFS_UTILS](https://github.com/NOAA-EMC/UFS_UTILS) project.
+
+This user guide is part of the <a href="../index.html">UFS_UTILS
+documentation</a>.
+
+## Where to find GFS GRIB2, NEMSIO and NetCDF data
```
GeorgeGayno-NOAA, line 17: Could you just link to the 'readthedocs' here?  If the 'rst' file changes, then the same change would have to be made here.

edwardhartnett: My intention is to go the other way around, so that the doxygen generates the readthedocs output automatically.

GeorgeGayno-NOAA (approved):
Branch builds with no doxygen warnings.  And the html looks correct.  I approve.  (Merge latest changes from 'develop').

PR #430, opened 2021-03-26
Add OpenMPI to Github Action

### LINKED ISSUE #425: Add OpenMPI testing
The two main flavors of MPI are MPICH and OpenMPI. We need to test with both, as we do on the ParallelIO project.

The reason is, we will have uses with both, and we need to be sure that we build and test cleanly on both.

kgerheiser:
Fixes #425 

This was more difficult than I thought it would be. When I first added OpenMPI I got undefined references like this:

```
ESMCI_XGrid_F.C:(.text._ZNK3MPI8Cartcomm5CloneEv[_ZNK3MPI8Cartcomm5CloneEv]+0x39): undefined reference to `MPI::Comm::Comm()'
2986
/usr/bin/ld: /home/runner/esmf/lib/libesmf.a(ESMCI_XGrid_F.o):ESMCI_XGrid_F.C:(.text._ZNK3MPI9Intercomm5MergeEb[_ZNK3MPI9Intercomm5MergeEb]+0x3d): more undefined references to `MPI::Comm::Comm()' follow
2987
/usr/bin/ld: /home/runner/esmf/lib/libesmf.a(ESMCI_XGrid_F.o):(.data.rel.ro._ZTVN3MPI8DatatypeE[_ZTVN3MPI8DatatypeE]+0x78): undefined reference to `MPI::Datatype::Free()'
```

The MPI C++ library wasn't being linked in for OpenMPI which is needed by ESMF. So, you have to enable `CXX` in CMake and for any target that links to ESMF needs to also link to `MPI::MPI_CXX`. I guess MPICH automatically links to its C++ library? 

Not sure why this is the case for OpenMPI. I haven't seen this anywhere else, but I don't think any NCEPLIBS use ESMF.


edwardhartnett:
@kgerheiser do you want to try and get this working in time for the 1.3.0 release?

kgerheiser:
Let me see.

edwardhartnett:
Can we get this working and merged soon?

kgerheiser:
Replaced by #503

PR #436, opened 2021-04-14
Unit test and bug fix for chgres_cube.fd/search_util.F90

### LINKED ISSUE #279: write unit tests for chgres_cube.fd/search_util.f90
OK, now it's time to move on to some more challenging testing. ;-)

We have this function in chgres_cube.fd/search_util.f90:

```
!> @brief Replace undefined surface values.
!!
!! Replace undefined values on the model grid with a valid value at
!! a nearby neighbor.  Undefined values are typically associated
!! with isolated islands where there is no source data.
!!
!! Routine searches a neighborhood with a radius of 100 grid points.
!! If no valid value is found, a default value is used.
!!
!! @note This routine works for one tile of a cubed sphere grid.  It
!! does not consider valid values at adjacent faces.  That is a 
!! future upgrade.
 subroutine search (field, mask, idim, jdim, tile, field_num, latitude, terrain_land, soilt_climo)

```

As a first step, can we add param tags in the doxygen to describe each of these parameters?

As a second step, write a test which constructs some parameters, calls this function, and checks the results. 

Seems like this code must be run with MPI, is that right @GeorgeGayno-NOAA ? If so, we can use the same techniques as the ParallelIO project to run the test with MPI. Basically:

```
# This is some custom CMake code from ParallelIO project.
include (LibMPI)

  add_mpi_test(pio_unit_test
    EXECUTABLE ${CMAKE_CURRENT_BINARY_DIR}/pio_unit_test
    NUMPROCS 4
    TIMEOUT ${DEFAULT_TEST_TIMEOUT})
```

I will see if I can get an empty test running with mpi to demonstrate.

### LINKED ISSUE #441: Possible bug in chgres_cube's sst_guess routine.
The `sst_guess` routine sets the SST values for small lakes that are far from other water bodies. The SST is based on latitude as follows:

```
if (latitude >= 60.0) then
   sst = 273.16
 elseif (abs(latitude) <= 30.0) then
   sst = 300.0
 else
   sst = (-0.8947) * abs(latitude) + 326.84
 endif
```

I suspect the first 'if' block should be `if (abs(latitude) >= 60)`. Otherwise, SST values south of minus 60 would be below freezing.

I don't think this is causing problems because there are likely very few isolated open water lakes south of minus 60. But it should be fixed anyway.

GeorgeGayno-NOAA:
Add a unit test for this module. 

Two types of search tests were added:

- First, a missing value is replaced with a valid neighboring value. 
- Second, a missing value is replaced by a default value. This is rare, but it can happen for an isolated island located far from a valid neighbor.

During testing, a bug was found when setting a default SST at points south of 60 degrees south latitude.

Fixes #279
Fixes #441


GeorgeGayno-NOAA:
Question, the search loop is threaded. How do you control the number of threads on Github? Should this test be run multiple times, using one thread, then multiple threads?

GeorgeGayno-NOAA:
> Question, the search loop is threaded. How do you control the number of threads on Github? Should this test be run multiple times, using one thread, then multiple threads?

I see you have an issue open for this.

GeorgeGayno-NOAA:
Hold off on any reviews. I want to add Kyle's recent changes for running mpi.

tests/chgres_cube/ftst_search_util.F90
```diff
@@ -0,0 +1,186 @@
+ program test_search_util
+
+! Test search_util using a simple 3x3 fv3 tile. Two types
+! of tests are performed. First, a missing value is 
+! replaced with a valid neighboring value. Second, a missing
+! value is replaced by a default value. This can happen
+! for an isolated island located far from a valid neighbor.
```
edwardhartnett, line 7: Excellent documentation here and throughout. It makes everything much more clear!

Please add your name here. Although we don't use doxygen, all tests should still include their author.

tests/CMakeLists.txt
```diff
@@ -4,6 +4,6 @@
 # Ed Hartnett 2/11/21
 
 # Add the test subdirecotries.
-add_subdirectory(chres_cube)
+add_subdirectory(chgres_cube)
```
edwardhartnett, line 7: Thanks for catching and correcting this!

edwardhartnett (approved):

PR #444, opened 2021-04-20
Minor documentation update in sorc/grid_tools.fd/regional_esg_grid.fd

edwardhartnett:
Part of #343 

In this PR I eliminate a stray "???" which should not be present.

Also added some author info that I learned from Rusty.


GeorgeGayno-NOAA (approved):
Looks good.

PR #448, opened 2021-04-21
Define constants in chgres_cube.fd/search_util.f90

### LINKED ISSUE #278: use of unexplained constants in chgres_cube code search_util.f90
In chgres_cure.fd/search_util.f90 I see:

```
subroutine sst_guess(latitude, sst)

 use esmf

 implicit none

 real(esmf_kind_r8), intent(in)  :: latitude

 real(esmf_kind_r8), intent(out) :: sst

 if (latitude >= 60.0) then
   sst = 273.16
 elseif (abs(latitude) <= 30.0) then
   sst = 300.0
 else
   sst = (-0.8947) * abs(latitude) + 326.84
 endif

 end subroutine sst_guess
```

We have a bunch of undocumented constants here. ;-)

These should be defined as constants and properly documented.

GeorgeGayno-NOAA:
Define unexplained constants.

Fixes #278


sorc/chgres_cube.fd/search_util.F90
```diff
@@ -215,27 +215,47 @@ subroutine search (field, mask, idim, jdim, tile, field_num, latitude, terrain_l
...
+!! Based loosely on the average annual SST
+!! values from ./fix_am/cfs_oi2sst1x1monclim19822001.grb
+!!
+!! The temperature in the polar and tropical regions
+!! is set to 273.16/300 Kelvin. Polar regions are
+!! poleward of 60 degrees. Tropical regions are within
+!! 30 degrees of the equator. In mid-latitudes, a
+!! linear change with latitude is used. 
+!!
+!! @param [in] latitude Latitude in degrees
```
edwardhartnett, line 229: Thanks for adding units! ;-) Always a good idea!!

Is this degrees and minutes or degrees and hundredths?

sorc/chgres_cube.fd/search_util.F90
```diff
@@ -215,27 +215,47 @@ subroutine search (field, mask, idim, jdim, tile, field_num, latitude, terrain_l
...
+!! @param [out] sst Default SST in Kelvin
 !! @author George Gayno NCEP/EMC
  subroutine sst_guess(latitude, sst)
 
  use esmf
 
  implicit none
 
+ real(esmf_kind_r8), parameter :: sst_polar_in_kelvin = 273.16 !< Default SST in polar
```
edwardhartnett, line 238: The usual convention is to name parameter variables in all caps, for example SST_POLAR_IN_KELVIN.

Of course since fortran ignores case anyway, this is purely a convention, and not a perfect one. However, it does help make clear in the code what part of a formula is a constant and what parts are variables.

sorc/chgres_cube.fd/search_util.F90
```diff
@@ -215,27 +215,47 @@ subroutine search (field, mask, idim, jdim, tile, field_num, latitude, terrain_l
...
  real(esmf_kind_r8), intent(out) :: sst
 
- if (abs(latitude) >= 60.0) then
-   sst = 273.16
- elseif (abs(latitude) <= 30.0) then
-   sst = 300.0
+ if (abs(latitude) >= polar_latitude) then
+   sst = sst_polar_in_kelvin
+ elseif (abs(latitude) <= tropical_latitude) then
+   sst = sst_tropical_in_kelvin
```
edwardhartnett, line 256: Thanks, this is a vast improvement in code readability!

sorc/chgres_cube.fd/search_util.F90
```diff
@@ -215,27 +215,47 @@ subroutine search (field, mask, idim, jdim, tile, field_num, latitude, terrain_l
...
-!> Set sst values based on latitude.
+!> Set default sst values based on latitude.
 !!
-!! @param latitude latitude input
-!! @param sst sst guess value to be set
+!! Based loosely on the average annual SST
+!! values from ./fix_am/cfs_oi2sst1x1monclim19822001.grb
+!!
+!! The temperature in the polar and tropical regions
+!! is set to 273.16/300 Kelvin. Polar regions are
```
edwardhartnett, line 224: Do you mean it is set to 273.16 in polar regions and 300 in tropical?

If so, please say so more clearly. If not, add even more clarity to this sentence. ;-)

sorc/chgres_cube.fd/search_util.F90
```diff
@@ -215,27 +215,47 @@ subroutine search (field, mask, idim, jdim, tile, field_num, latitude, terrain_l
 
  end subroutine search
 
-!> Set sst values based on latitude.
+!> Set default sst values based on latitude.
```
edwardhartnett, line 218: When using SST as an acronym, please capitalize it for clarity.

Also, the first time using an acronym in a documentation block (i.e. the first sentence in this case), define it:

Set default Sea Surface Temperatures (SST) values based on latitude.

edwardhartnett (approved):

PR #452, opened 2021-04-23
Update FindESMF.cmake

aerorahul:
This PR updates `FindESMF.cmake` based on `esmf.mk` to define the following variables:
- create `ESMF_VERSION` string based on `ESMF_VERSION_STRING_GIT`
- define `MAJOR`, `MINOR`, `PATCH`, `TWEAK` components of `ESMF_VERSION`
- add a variable `ESMF_BETA_RELEASE` to indicate returned version is a Beta release.

Usage:
```
find_package(ESMF <VERSION> <REQUIRED>)
```

If the found version is less than the desired `VERSION`, `find_package` will fail with an error message.

I have tested a stripped down build on my macOS and Orion, but not on the various platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ endif()
 # Find packages.
 find_package(NetCDF REQUIRED C Fortran)
 find_package(MPI REQUIRED C Fortran)
-find_package(ESMF MODULE REQUIRED)
+find_package(ESMF 8.1.0.27 REQUIRED)
 
 if(OPENMP)
   find_package(OpenMP REQUIRED COMPONENTS Fortran)
--- a/cmake/FindESMF.cmake
+++ b/cmake/FindESMF.cmake
@@ -72,6 +72,20 @@ if (ESMF_FOUND)
     endif()
   endforeach()
 
+  # Construct ESMF_VERSION from ESMF_VERSION_STRING_GIT
+  if(ESMF_FOUND)
+    # ESMF_VERSION_MAJOR and ESMF_VERSION_MINOR are defined in ESMFMKFILE
+    set(ESMF_VERSION 0)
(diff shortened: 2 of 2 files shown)


edwardhartnett:
These changes look great. CI is failing due to the macos homebrew problem... @kgerheiser does that get fixed with a brew update command somewhere?

edwardhartnett:
OK, turns out there's a problem here. See the comments in this issue:
https://github.com/NOAA-EMC/hpc-stack/issues/208

kgerheiser:
Yes, that's a problem, but the good news is since we have our own `FindESMF.cmake` we can do whatever we like regarding version checking. Might have to remove the version from `find_package_handle_standard_args` and handle it ourselves or otherwise modify it before passing it.

aerorahul:
> Yes, that's a problem, but the good news is since we have our own `FindESMF.cmake` we can do whatever we like regarding version checking. Might have to remove the version from `find_package_handle_standard_args` and handle it ourselves or otherwise modify it before passing it.

@kgerheiser 
Do you want to implement the subtraction and handling of the `TWEAK`?  
If not, I will have to get back to this sometime next week.

I would prefer keeping the `find_package_handle_standard_args` and work to find a solution that fits in.

edwardhartnett:
OK, last week the ESMF team did a release for us.

Perhaps we should forget about checking for beta releases, and only support checking for full releases in the CMake.

This would match what is generally done. For example, when requiring a netcdf-c version, we will only worry about actual netcdf-c releases, not any intermediate releases. Dealing with those will be up to the developers, not hpc-stack.

With this relaxation, the problem becomes a lot simpler, right?

kgerheiser:
@edwardhartnett Yes, just getting the version if straightforward.

Issue #461, opened 2021-04-26
Put rest of chgres_cube code into modules


kgerheiser:
`utils.F90` and `write_data.F90` don't use modules. It's good for all code to be in modules for explicit type-checking of functions.

GeorgeGayno-NOAA:
write_data.F90 was placed in a module under this PR - #631

LarissaReames:
@kgerheiser Are there any other thoughts for parts of the code that could be split into separate modules?

GeorgeGayno-NOAA:
`utils.F90` and `write_data.F90` are now modules. Closing.

Issue #460, opened 2021-04-26
Split input_data.F90 into `atm_input_data`, `sfc_input_data`, and `nst_target_data`


kgerheiser:
`input_data.F90` is a really long file that I think could be broken up to make it more managable. It seems like it does 3 different things for 3 types of data, and it would make sense to group them together.

GeorgeGayno-NOAA:
@kgerheiser This will make writing unit tests easier. What is the status? I can work on this if you are busy with other tasks.

kgerheiser:
I'll put in a PR today.

LarissaReames:
@kgerheiser did you ever go back to try to tackle this? I think it's still worth doing as that file really is quite large.

GeorgeGayno-NOAA:
> @kgerheiser did you ever go back to try to tackle this? I think it's still worth doing as that file really is quite large.

Unfortunately, Kyle no longer works for NOAA. But I still think his suggestion is a good one.

LarissaReames:
> > @kgerheiser did you ever go back to try to tackle this? I think it's still worth doing as that file really is quite large.
> 
> Unfortunately, Kyle no longer works for NOAA. But I still think his suggestion is a good one.

Yes I think so too. Any thoughts on the best way to do this? Do we just do a straight up split of those subroutines in to 3 files?

GeorgeGayno-NOAA:
> > > @kgerheiser did you ever go back to try to tackle this? I think it's still worth doing as that file really is quite large.
> > 
> > 
> > Unfortunately, Kyle no longer works for NOAA. But I still think his suggestion is a good one.
> 
> Yes I think so too. Any thoughts on the best way to do this? Do we just do a straight up split of those subroutines in to 3 files?

Maybe the first step is to do an inventory of all the routines and what they do. Not all of the routines read data. Some could probably be moved to the utility module.

Issue #449, opened 2021-04-22
some packages required in CMakeLists.txt which are no longer needed?


edwardhartnett:
I think the list of required packages in CMakeLists.txt contains some packages that are no longer needed:

```
find_package(gfsio REQUIRED)
find_package(sfcio REQUIRED)
find_package(w3nco REQUIRED)
find_package(bacio REQUIRED)
find_package(nemsio REQUIRED)
find_package(sigio REQUIRED)
find_package(sp REQUIRED)
find_package(ip REQUIRED)
find_package(w3emc REQUIRED)
find_package(g2 REQUIRED)
find_package(wgrib2 REQUIRED)
```

w3emc does not seem to be used. Haven't we eliminated use of some of the others as well?

GeorgeGayno-NOAA:
Good catch. gfsio is not used either.

Recall we eliminated a few libraries when global_chgres was removed from UFS_UTILS:
https://github.com/NOAA-EMC/UFS_UTILS/pull/299#pullrequestreview-596742674

GeorgeGayno-NOAA:
Can I fold this in with #248 in one pull request?

PR #462, opened 2021-04-26
Cleanup/chgres cube modules

kgerheiser:
Making this a draft PR to start, but I ran the regression tests and all passed.

In this PR:

* Add the remaining code into modules (#461), and this leads to a circular dependency between write_data and atomsphere.F90/surface.F90. I move out the target fields into their own module similar to how `input_data.F90` contains all the `_input_grid` fields.

* Split `input_data.F90` into 3 modules: `atm_input_data.mo`, `sfc_input_data_mod`, `nst_input_data_mod`, to make `input_data.F90` more managable. (#460)

* Remove `find_package(w3emc)` and `find_package(gfsio)` (#449)

Thoughts on this?


GeorgeGayno-NOAA:
> Making this a draft PR to start, but I ran the regression tests and all passed.
> 
> In this PR:
> 
> * Add the remaining code into modules (#461), and this leads to a circular dependency between write_data and atomsphere.F90/surface.F90. I move out the target fields into their own module similar to how `input_data.F90` contains all the `_input_grid` fields.
> * Split `input_data.F90` into 3 modules: `atm_input_data.mo`, `sfc_input_data_mod`, `nst_input_data_mod`, to make the files more managable. (#460)
> * Remove `find_package(w3emc)` and `find_package(gfsio)` (#449)

As an FYI, I was planning to remove the unused packages under my own branch. If you want to do that here, then remove those libraries from all the build modules and update the README.md file. I like what you are doing, but you may be trying to do too many things under one PR. 

> 
> Thoughts on this?

kgerheiser:
Fair enough. I will split it up into 2 PRs. You can do the find_package thing.

PR #466, opened 2021-04-27
Add version checking for each library

### LINKED ISSUE #443: check versions of all dependent packages
When we look for dependent packages in the CMake build we must know what versions of each package are acceptable. That is, what version of ESMF is required to run the code.

kgerheiser:
Fixes #443 and supersedes #452

I took out the tweak version for the beta and added a message if using a beta release of ESMF.

I put `4.7.3` for NetCDF because the CI uses `4.7.3`, though may want to relax more.

The versions for NCEPLIBS were chosen based upon their first CMake release.

I put ESMF at 8.0.0.


cmake/FindESMF.cmake
```diff
@@ -72,6 +72,19 @@ if (ESMF_FOUND)
...
 
+  # Construct ESMF_VERSION from ESMF_VERSION_STRING_GIT
+  if(ESMF_FOUND)
+    # ESMF_VERSION_MAJOR and ESMF_VERSION_MINOR are defined in ESMFMKFILE
+    set(ESMF_VERSION 0)
+    set(ESMF_VERSION_PATCH ${ESMF_VERSION_REVISION})
+        set(ESMF_BETA_RELEASE FALSE)
+    if(ESMF_VERSION_BETASNAPSHOT MATCHES "^('T')$")
+      set(ESMF_BETA_RELEASE TRUE)
+      message(STATUS "Beta release of ESMF detected.")
```
aerorahul, line 83: It would still be good to echo what the beta snapshot it detected.
Perhaps something like:
```
string(REGEX REPLACE ".*beta_snapshot_*\([0-9]*\).*" "\\1" ESMF_BETA_SNAPSHOT "${ESMF_VERSION_STRING_GIT}")
message(STATUS "Detected ESMF Beta snapshot ${ESMF_BETA_SNAPSHOT}")
```

CMakeLists.txt
```diff
@@ -45,25 +45,23 @@ if(CMAKE_C_COMPILER_ID MATCHES "^(Intel)$")
 endif()
 
 # Find packages.
-find_package(NetCDF REQUIRED C Fortran)
+find_package(NetCDF 4.7.3 REQUIRED C Fortran)
 find_package(MPI REQUIRED C Fortran)
-find_package(ESMF MODULE REQUIRED)
+find_package(ESMF 8.0.0 MODULE REQUIRED)
```
aerorahul, line 50: `MODULE` is not required.

aerorahul (changes requested):
see comments.

edwardhartnett (approved):

aerorahul (approved):
One minor fix.
The rest looks good.

Please also open a PR in `CMakeModules` and then a following in the UFS-weather-model.

GeorgeGayno-NOAA (changes requested):
It won't compile for me on WCOSS-Cray. Recall, the hpc-stack is not installed on that machine.

kgerheiser:
Can you give the exact error?

GeorgeGayno-NOAA:
> Can you give the exact error? None of this should be dependent on hpc-stack.

We point to a older version of netcdf:

```
CMake Error at /gpfs/hps/usrx/local/prod/cmake/3.16.2/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find NetCDF: Found unsuitable version "4.3.3.1", but required is
  at least "4.7.3" (found /opt/cray/netcdf/4.3.3.1/INTEL/14.0/include)

```

kgerheiser:
The weather model uses NetCDF 4.7.4 on WCOSS Cray, so we could use that. Or relax the restriction on the version.

kgerheiser:
It obviously works with NetCDF 4.3.3, so I think we should relax the version.

aerorahul:
ugh cray!

CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "^(Intel)$")
 endif()
 
 # Find packages.
-find_package(NetCDF 4.7.3 REQUIRED C Fortran)
+find_package(NetCDF 4.3.3 REQUIRED C Fortran)
```
edwardhartnett, line 48: Actually I don't know that we should do this. We are not testing with 4.3.3. We really don't know that this code works with that version, because we don't have much testing anyway.

We have best confidence that it works with 4.7.4. I think we should require that.

kgerheiser:
I kept having issues on WCOSS Cray. I tried using NetCDf 4.7.4, but somehow 4.3.3 was still being pulled in. I traced it back to ESMF which was also linking in the NetCDF it was built with (4.3.3).  If we get ESMF 8.1.1 installed on WCOSS Cray as part of https://github.com/NOAA-EMC/hpc-stack/pull/222#issuecomment-828625456 then everything will be consistent and we can use that. I don't want to point to the beta snapshot the weather model is using when it's about to be replace with 8.1.1.

edwardhartnett:
OK, how about we change the version of netcdf back to 4.3.3 for this PR, if that will allow us to merge it.

Then we can circle back around to find the problem on the Cray.

kgerheiser:
@GeorgeGayno-NOAA this has been updated so it works on WCOSS Cray

GeorgeGayno-NOAA (approved):

GeorgeGayno-NOAA:
Commit f6eda81 does not compile on WCOSS-Cray (the previous commit to develop does compile). I get this error:

```
CMake Error at cmake/FindNetCDF.cmake:241 (add_library):
  add_library called with IMPORTED argument but no library type.
Call Stack (most recent call first):
  CMakeLists.txt:48 (find_package)


CMake Error at cmake/FindNetCDF.cmake:242 (set_target_properties):
  set_target_properties Can not find target to add properties to:
  NetCDF::NetCDF_C
Call Stack (most recent call first):
  CMakeLists.txt:48 (find_package)


CMake Error at cmake/FindNetCDF.cmake:241 (add_library):
  add_library called with IMPORTED argument but no library type.
Call Stack (most recent call first):
  CMakeLists.txt:48 (find_package)


CMake Error at cmake/FindNetCDF.cmake:242 (set_target_properties):
  set_target_properties Can not find target to add properties to:
  NetCDF::NetCDF_Fortran
Call Stack (most recent call first):
  CMakeLists.txt:48 (find_package)


CMake Error at cmake/FindNetCDF.cmake:247 (target_link_libraries):
  Cannot specify link libraries for target "NetCDF::NetCDF_Fortran" which is
  not built by this project.
Call Stack (most recent call first):
  CMakeLists.txt:48 (find_package)
```

aerorahul:
@edwardhartnett @kgerheiser 
Please open the PR from the changes in `FindESMF.cmake` that went in this PR into the CMakeModules repository.
CMakeModules is used by other projects.

kgerheiser:
Opened a PR

Issue #463, opened 2021-04-27
develop fails to compile with gfortran10 due to missing -fallow-argument-mismatch compile flag


edwardhartnett:
For some reason the develop branch is not compiling for me today.

What am I doing wrong? Do I need a more recent version of one of my libraries?

```
cd /home/ed/UFS_UTILS/b/sorc/chgres_cube.fd && /usr/bin/gfortran  -I/usr/local/NCEPLIBS-nemsio-2.5.1/include -I/usr/local/NCEPLIBS-sfcio-1.4.0/include -I/usr/local/NCEPLIBS-sigio-2.3.0/include -I/usr/local/NCEPLIBS-bacio-2.4.1/include_4 -I/usr/local/NCEPLIBS-sp-2.3.3/include_d -I/usr/local/w3nco-2.4.0/include_d -I/home/ed/esmf/mod -I/home/ed/esmf/include -I/usr/local/netcdf-fortran-4.5.3_netcdf-c-4.7.4/include -I/usr/include -I/usr/local/NCEPLIBS-wgrib2-2.9.0/include -I/usr/local/include  -g -fbacktrace -ffree-line-length-0 -fdefault-real-8 -fconvert=big-endian -O3 -Jmod   -c /home/ed/UFS_UTILS/sorc/chgres_cube.fd/model_grid.F90 -o CMakeFiles/chgres_cube_lib.dir/model_grid.F90.o
/home/ed/UFS_UTILS/sorc/chgres_cube.fd/model_grid.F90:1061:22:

  855 |   call MPI_BCAST(temp_num,10,MPI_CHAR,0,MPI_COMM_WORLD,error)
      |                 2     
......
 1061 |        call MPI_BCAST(dx,1,MPI_REAL8,0,MPI_COMM_WORLD,error)
      |                      1
Error: Type mismatch between actual argument at (1) and actual argument at (2) (REAL(8)/CHARACTER(*)).
make[2]: *** [sorc/chgres_cube.fd/CMakeFiles/chgres_cube_lib.dir/build.make:119: sorc/chgres_cube.fd/CMakeFiles/chgres_cube_lib.dir/model_grid.F90.o] Error 1
make[2]: Leaving directory '/home/ed/UFS_UTILS/b'
make[1]: *** [CMakeFiles/Makefile2:1913: sorc/chgres_cube.fd/CMakeFiles/chgres_cube_lib.dir/all] Error 2
[...]

PR #469, opened 2021-04-28
Turn on fortran flag -fallow-argument-mismatch in chgres_cube for gfortran10

edwardhartnett:
Part of #463


sorc/chgres_cube.fd/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ elseif(CMAKE_Fortran_COMPILER_ID MATCHES "^(GNU)$")
   set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -ffree-line-length-0 -fdefault-real-8 -fconvert=big-endian")
 endif()
 
+# Turn on this argument mismatch flag for gfortran10.
```
kgerheiser, line 27: This needs to go in side the check for the GNU compiler. You don't want to add those options for any compiler.

kgerheiser (changes requested):

sorc/chgres_cube.fd/CMakeLists.txt
edwardhartnett: Duh. I have fixed.

kgerheiser (approved):

PR #474, opened 2021-04-28
Re-add NETCDF env variable so CMake can find NetCDF on WCOSS Cray

### LINKED ISSUE #475: NetCDF not found on WCOSS Cray
`NETCDF` env variable was removed with latest commit because I was testing with NetCDF 4.7.4 which sets the variable automatically and it was never re-added.

kgerheiser:
Variable was removed by accident

Fixes #475


kgerheiser:
CMake was already found and the location cached. So I didn't notice that it didn't find it from a fresh build.

GeorgeGayno-NOAA (approved):
It works! Thanks.

Issue #467, opened 2021-04-27
Update testing to use netcdf-c-4.7.4


edwardhartnett:
We should be testing with whatever atmfv3 uses.

edwardhartnett:
I started on this.

We need to build HDF5, netcdf-c, and netcdf-fortran from source and cache them, so we can test against various versions and flavors other than those available on ubuntu, which is usually behind the actual release. For example, netcdf-c-4.8.0 is out now.

So I will add testing for building with HDF5 1.10.7 and netcdf-c-4.7.4.

@kgerheiser what's not clear is whether this can be merged with our master build, or whether we will have to use multiple workflow files.

PR #492, opened 2021-05-03
Adding testing with custom-build netcdf and HDF5 versions.

edwardhartnett:
Part of #467 

Adding testing with custom-build netcdf and HDF5 versions.


kgerheiser:
This will be helpful for the Intel build

edwardhartnett:
@kgerheiser this works now.

Should it be left as a separate workflow? Or do you want to try merging it into the existing workflow?

(There are other versions of netcdf we may check against, as well as perhaps multiple versions of ESMF we might want to test against.)

.github/workflows/netcdf-4.7.4.yml
```diff
@@ -0,0 +1,212 @@
...
+        python3 -m pip install gcovr
+
+    - name: cache-esmf
+      id: cache-esmf
+      uses: actions/cache@v2
+      with:
+        path: ~/esmf
+        key: esmf-${{ runner.os }}-8.0.1-2
+
+    - name: build-esmf
```
kgerheiser, line 68: Build NetCDF should come before building ESMF. If ESMF weren't cached it would fail building.

kgerheiser (changes requested):

.github/workflows/netcdf-4.7.4.yml
edwardhartnett: Good catch! I have fixed, and pointed ESMF at the installed netCDF build.

edwardhartnett:
@kgerheiser please approve and we can merge.

kgerheiser:
What do you think of putting this into the existing workflow instead of creating a new one? I am currently testing it on a branch.

edwardhartnett:
Fine if we can.

But I would like to add testing for netcdf-c-4.8.0 as well. ;-) Can you do that at the same time?

It builds with the same version of Fortran as 4.7.4.

kgerheiser:
Sure, adding NetCDF 4.8.0 is easy. I'll add it as a matrix.

kgerheiser:
See #503

PR #503, opened 2021-05-06
Github Action Improvements

### LINKED ISSUE #425: Add OpenMPI testing

### LINKED ISSUE #467: Update testing to use netcdf-c-4.7.4

### LINKED ISSUE #494: Set up CI testing with free intel fortran compiler
I'm reading on the interwebs that intel has a free version of its fortran compiler for non-commercial use on Linux. Well, that's us!

So it seems like we could use this to get Intel testing as well as GNU. Perhaps we can do this on GitHub actions, perhaps on private Jenkins servers.

### LINKED ISSUE #501: use ctest options --rerun-failed --output-on-failure in CI runs
@kgerheiser this is super helpful.

Using ctest with these options shows the output of a failing test in the CI. Otherwise, all we know is that a test failed, and there's no way to see the output.

We should be using these options everywhere...

kgerheiser:
Add Intel OneAPI build to CI: Fix #494 

Add arguments to print on error for ctest: Fix #501 

Add OpenMPI build to CI: Fix #425

Build NetCDf 4.7.4 in CI: Fix #467 and replace #492

Upgrade to Jasper 2.0.25. Jasper 2.0.22 has a known bug with Intel compilers.

Re-name yaml files. I think they were too long and the full name didn't show up in the Github UI.

I have it setup so that all the variants (mac/linux, nceplibs develop/1.3.0, openmpi/mpich) run in the one script, while the other is for Intel. Now this can broken up in a number of ways, but I did it like this to reduce code duplication. Ideally you could use JSON anchors or some other mechanism to re-use steps, but it's not currently possible.


edwardhartnett:
Awesome! 

Does UFS_UTILS use jasper?

.github/workflows/intel-oneapi.yml
```diff
@@ -0,0 +1,167 @@
...
+      uses: actions/checkout@v2
+      with: 
+        path: ufs_utils
+
+    - name: build
+      run: |
+        export ESMFMKFILE=~/esmf/lib/esmf.mk
+        cd ufs_utils
+        mkdir build && cd build
+        cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH='~;~/jasper;~/nceplibs;~/netcdf' -DENABLE_DOCS=On
```
edwardhartnett, line 159: We only need to ENABLE_DOCS on one build, the GNU build I would suggest. It adds a lot of output which makes scrolling through it a bit of a pain...

.github/workflows/intel-oneapi.yml
```diff
@@ -0,0 +1,167 @@
...
+        make install
+           
+    - name: build-netcdf-c
+      if: steps.cache-netcdf.outputs.cache-hit != 'true'
+      run: |
+        export CC=mpiicc
+        export CPPFLAGS=-I${HOME}/netcdf/include
+        export LDFLAGS=-L${HOME}/netcdf/lib
+        wget https://www.unidata.ucar.edu/downloads/netcdf/ftp/netcdf-c-4.7.4.tar.gz &> /dev/null
+        tar -xzf netcdf-c-4.7.4.tar.gz
```
edwardhartnett, line 58: In a future PR are you going to matrix the netcdf-c version, so we can test with 4.7.4 and also 4.8.0?

.github/workflows/intel-oneapi.yml
```diff
@@ -0,0 +1,167 @@
...
+      id: cache-netcdf
+      uses: actions/cache@v2
+      with:
+        path: ~/netcdf
+        key: netcdf-c-${{ runner.os }}-4.7.4-intel
+
+    - name: build-hdf5
+      if: steps.cache-netcdf.outputs.cache-hit != 'true'
+      run: |
+        export CC=mpiicc
```
edwardhartnett, line 43: I like to use "set -x" at the top of all my build commands, but I see you have removed them. I wonder if there's a way we could turn that one in one place for all commands...

edwardhartnett (approved):

edwardhartnett (approved):

.github/workflows/intel-oneapi.yml
kgerheiser: I added it. You can invoke the shell with custom commands. I had to do this with Intel to load the environment with a login shell (`-l`) so that it sourced `.bash_profile`. Here I just added `-x`.

.github/workflows/intel-oneapi.yml
kgerheiser: I added this. In the end I made it a separate workflow.

kgerheiser:
@GeorgeGayno-NOAA or @edwardhartnett can you disable the `build` status check and then enable the new ones (or maybe merge first, then enable)?

edwardhartnett:
@kgerheiser I have given you admin rights on this repo, so you can set the branch protection.

PR #509, opened 2021-05-07
Feature/soiltype fix

LarissaReames-NOAA:
Addresses Issue #246 

This fix took a while to figure out, but ended up being rather simple. 

**_Problem:_** When regridding soil type to the target grid (instead of using climatology), some grid points over land ice had 0 soil type (which made RUC LSM throw a fit). 

**_Solution:_** Create a separate set of regridding routines on to the target grid for sotyp_from_climo=.false. over both land ice and non-land ice. What I'd failed to recognize in past attempts to fix this was the need for the zeroregion=ESMF_REGION_SELECT flag in the second regridding routine which is over land but not land ice points. Without this flag, the soil type values that had been properly re-gridded at land ice points were being re-set to zero. That has now been resolved and the target grid no longer contains points with zero soil moisture at land ice points. 

**_Testing:_** These changes can't really be tested by being separated from the rest of the interp routine in its current form. Testing for these changes will need to be made possible  in future surface.F90 work. 

The resulting files no longer cause fatal RUC LSM errors during weather model initialization.


edwardhartnett:
How can this code be tested? That is, what would the test look like for this subroutine?

LarissaReames-NOAA:
@edwardhartnett A test for the entire interp subroutine would probably involve several different calls to the subroutine, each time sending it small test data sets that look like data that would come from various sources. One test call wouldn't be enough as there are various statements that are called depending on namelist entries (e.g., input source, whether to use various climatological fields on the target grid, etc.). 

If we're only looking at the exact code I added, the specific test would involve setting up a small ESMF input and target grid containing various surface fields that look like what comes from HRRR data and using flags that result in soil and vegetation type from the input grid being used on the target grid.

edwardhartnett:
Can you add the test you describe in your second paragraph?

LarissaReames-NOAA:
@edwardhartnett That's going to take quite a bit of work. One question I have that's relevant: If all I'm testing here is the interp subroutine, can I make use of other subroutines in the repository? Writing up a test for this is going to require recreating all of the surface input and target fields, and re-writing all of that code would seem really redundant. Would I be able to call the subroutines that create these fields to cut down on repeated code? Otherwise I'd basically be copy and pasting it in.

GeorgeGayno-NOAA:
@LarissaReames-NOAA I still don't understand why these "xxxx_from_climo" options require extra logic to work. Is there something wrong with the input soil type data? I don't encounter these issues when I use the pre-generated surface data created by the grid generation step. Can your desired surface data be made part of the grid generation where there are checks to ensure the data matches the model land mask? And where there is a consistency check for land ice points? My guess is that might solve this problem. A benefit would be to make these datasets available to other groups. And since fields such as soil type don't change unless the grid changes, you will eliminate a redundant mapping every time chgres is run.

JeffBeck-NOAA:
> @LarissaReames-NOAA I still don't understand why these "xxxx_from_climo" options require extra logic to work. Is there something wrong with the input soil type data? I don't encounter these issues when I use the pre-generated surface data created by the grid generation step. Can your desired surface data be made part of the grid generation where there are checks to ensure the data matches the model land mask? And where there is a consistency check for land ice points? My guess is that might solve this problem. A benefit would be to make these datasets available to other groups. And since fields such as soil type don't change unless the grid changes, you will eliminate a redundant mapping every time chgres is run.

@GeorgeGayno-NOAA, the problem addressed here was unrelated to external model data soil type, but rather an unresolved issue with land ice points.  The surface data coming from the RAP and HRRR models are (for now) in a format that is incompatible with the sfc_climo code, and would need to be modified for that code to read it in.  There are also a number of discrepancies between WRF-based surface data and the FV3 target grid/land mask that require these extra routines to handle interpolation correctly.

LarissaReames-NOAA:
Just to clarify, really the main problem these ****_from_climo variables address is that GSL wants these exact surface variables to be as close to how they are in HRRR as possible, and the data to create those arrays aren't presently available in a format that the surface climo array can process. The updates I made here don't really address any real "discrepancies" between how HRRR/WRF and FV3 handle land mask. This PR is really all about correcting my improper implementation of regridding soil type to the target grid for writing to file, with the main updates being interpolating at land ice points (which is not done for soil_type_from_input_grid) and including the zeroregion option when regridding on non-land ice points since that's a second regridding action on that field.

LarissaReames-NOAA:
I've written a test that passes on most platforms, but it appears there's something causing the array returned from the tested subroutine (interp) to be slightly different on Mac. There also appear to be memory leaks associated with the ESMF routines that I'm utilizing (I've removed all the ones associated with my actual code), and I'm not really sure what to do about that.

kgerheiser:
Could you remove `-fsanitize=address` from the top-level CMakeLists.txt?

LarissaReames-NOAA:
> Could you remove `-fsanitize=address` from the top-level CMakeLists.txt?

@edwardhartnett  added that early last week, so I'll let him respond on what to do about this.

LarissaReames-NOAA:
After adding in the LSAN file to suppress memory leak traces associated with ESMF, and building ESMF with debug and leak sanitizer flags, this branch now builds without errors and passes all unit tests.

LarissaReames-NOAA:
Well, all tests *were* passing, but it looks like there's some sort of memory issue with FieldBundle (in the nst_landfill test) that is only showing up now with a version of ESMF compiled in debug mode with sanitizers turned on. I tried adding in an ASAN file to suppress address warnings in the esmf library, but I was unsuccessful.

LarissaReames-NOAA:
Well, we haven't heard back from ESMF in about 2 weeks on some of the problems that continue to plague the new tests in this branch. What kind of strategies do we have moving forward since it's clear that these leaks are coming from an external library?

JeffBeck-NOAA (approved):
Tested on Hera.  All consistency checks pass except the two that require baseline changes after this PR is merged (this is expected, since the bug was present in the code used to produce the baseline data).  All unit tests pass as well.  Approving.

build_all.sh
```diff
@@ -8,11 +8,9 @@
 
 set -eux
 
-#compiler=gnu
-
 target=${target:-"NULL"}
```
GeorgeGayno-NOAA, line 11: Hold on. You changed the default to use GNU. AND to build using the debug option.

LarissaReames-NOAA: Gah, that was never meant to make it in to a commit. My apologies. I'll open a PR to fix that momentarily.

GeorgeGayno-NOAA:
> Tested on Hera. All consistency checks pass except the two that require baseline changes after this PR is merged (this is expected, since the bug was present in the code used to produce the baseline data). All unit tests pass as well. Approving.

Where is the new baseline data for the tests that were failing? I need to move it to all machines.

LarissaReames-NOAA:
> > Tested on Hera. All consistency checks pass except the two that require baseline changes after this PR is merged (this is expected, since the bug was present in the code used to produce the baseline data). All unit tests pass as well. Approving.
> 
> Where is the new baseline data for the tests that were failing? I need to move it to all machines.

Working on those right now. I'll drop in the location when they're ready.

LarissaReames-NOAA:
So It looks like there's only one baseline that needs to be changed. I think Jeff had an additional failure due to something else, perhaps the incorrect gnu compiler push. There are no points in the 13km conus RAP baseline file with soil type of 0 at land points, the problem this PR fixes, so there shouldn't be any failures. I'm not getting failure on that test with the current develop branch, so the only baseline that needs to be updated is the 3km_conus_hrrr_newsfc test, and those can be found here: 

Hera: `/scratch1/BMC/gsd-fv3/Larissa.Reames//reg-tests/chgres-cube/3km_conus_hrrr_newsfc_grib2`

Jet: `/lfs4/NAGAPE/hpc-wof1/lreames/reg-tests/chgres-cube/3km_conus_hrrr_newsfc_grib2`

 Note only the sfc file has changed, as is expected.

GeorgeGayno-NOAA:
> So It looks like there's only one baseline that needs to be changed. I think Jeff had an additional failure due to something else, perhaps the incorrect gnu compiler push. There are no points in the 13km conus RAP baseline file with soil type of 0 at land points, the problem this PR fixes, so there shouldn't be any failures. I'm not getting failure on that test with the current develop branch, so the only baseline that needs to be updated is the 3km_conus_hrrr_newsfc test, and those can be found here:
> 
> Hera: `/scratch1/BMC/gsd-fv3/Larissa.Reames//reg-tests/chgres-cube/3km_conus_hrrr_newsfc_grib2`
> 
> Jet: `/lfs4/NAGAPE/hpc-wof1/lreames/reg-tests/chgres-cube/3km_conus_hrrr_newsfc_grib2`
> 
> Note only the sfc file has changed, as is expected.

Thanks. I will run a quick test on all machines.

build_all.sh
GeorgeGayno-NOAA: > Hold on. You changed the default to use GNU. AND to build using the debug option.

My caps lock got stuck. Did not mean to yell.

GeorgeGayno-NOAA:
> > So It looks like there's only one baseline that needs to be changed. I think Jeff had an additional failure due to something else, perhaps the incorrect gnu compiler push. There are no points in the 13km conus RAP baseline file with soil type of 0 at land points, the problem this PR fixes, so there shouldn't be any failures. I'm not getting failure on that test with the current develop branch, so the only baseline that needs to be updated is the 3km_conus_hrrr_newsfc test, and those can be found here:
> > Hera: `/scratch1/BMC/gsd-fv3/Larissa.Reames//reg-tests/chgres-cube/3km_conus_hrrr_newsfc_grib2`
> > Jet: `/lfs4/NAGAPE/hpc-wof1/lreames/reg-tests/chgres-cube/3km_conus_hrrr_newsfc_grib2`
> > Note only the sfc file has changed, as is expected.
> 
> Thanks. I will run a quick test on all machines.

I updated the baseline data on all machines. All regression tests now pass.

JeffBeck-NOAA:
Great news!  Thanks, @GeorgeGayno-NOAA!

Issue #516, opened 2021-05-18
General cleanup of surface.F90 needed


LarissaReames-NOAA:
Similar to Issue #504 , the entire file surface.F90 could use some clean up. Outside of the interp subroutine, there's some old code in there that's no longer used, and an opportunity in nst_land_fill to move to using a loop over a FieldBundle to do most of the work of that subroutine. As this won't involve adding any functionalities that aren't already there, but mostly will be removing old code, I don't anticipate many tests being needed. A short test could probably be used to check that the newly-modified nst_land_fill is properly filling land points with the expected values.

PR #517, opened 2021-05-19
Clean up surface.F90 and simplify nst_land_fill with FieldBundle loop

LarissaReames-NOAA:
This PR addresses Issue #516 


LarissaReames-NOAA:
I'll make this PR again shortly. It looks like I built these changes on top of the wrong branch. I'll re-build on top of develop and re-submit.

PR #518, opened 2021-05-19
Clean up surface.F90 and simplify nst_land_fill with FieldBundle loop

LarissaReames-NOAA:
This PR addresses Issue #516


edwardhartnett:
Removing unused code is a great idea and lightens the load of testing, documenting, and understanding the code!

PR #252, opened 2021-02-01
release/public-v2 chgres_cube: Add regression tests for new regional chgres_cube features

LarissaReames:
This PR adds new regression tests for various combinations of the added regional functionality in chgres_cube.

Drivers for Jet and Hera have been updated. I'll need assistance from someone with access to the other platforms to update those drivers using my new drivers as a guideline. @GeorgeGayno-NOAA has previously offered to help with this. 

My baseline files, along with the corresponding grid and fix files, will need to be put in the same location as the others. I've been informed that the RRFS grids are still being updated as I write this, so these files may not be 100% ready to be copied, but that should change soon. George's help will be needed here as well. The locations of my files are/will be:

Hera: /scratch1/BMC/gsd-fv3/Larissa.Reames/chgres_cube/reg_tests.public.v2/
Jet: /home/Larissa.Reames/hpc-wof1/chgres_cube/reg_tests.public.v2

I will work on modifying these new tests/driver scripts to work with the develop branch in the coming weeks. 

This PR addresses Issue #181

PR #519, opened 2021-05-19
Clean up surface.F90 and simplify nst_land_fill with FieldBundle loop (this one will be correct, I swear)

LarissaReames-NOAA:
This PR addresses Issue #516

Third time's a charm


GeorgeGayno-NOAA:
@LarissaReames-NOAA Why create the field bundle in 'nst_land_fill'? Why not do that earlier in say 'create_nst_esmf_fields'. That way the bundle can be used in other places. For example, in routine 'interp' a bundle regrid could be done to shorten the code by eliminating all those individual regrids.

LarissaReames-NOAA:
> @LarissaReames-NOAA Why create the field bundle in 'nst_land_fill'? Why not do that earlier in say 'create_nst_esmf_fields'. That way the bundle can be used in other places. For example, in routine 'interp' a bundle regrid could be done to shorten the code by eliminating all those individual regrids.

The fieldbundle used here actually doesn't include **all** of the nst fields as there are two (tref and xz) which need to be set to different values then the others which get set to 0. It was either that, or include those two in the bundle write up an if statement that checked for the name of the field and did different things based on that. I figure it's probably close to a wash in code length difference between those two approaches. 

I actually have the latter coded up in a different branch, but I'm waiting for the soiltype fix to be merged in before submitting that PR. In that branch, I use bundles for each type of masked regridding and also use those same bundles to loop over for the search function as well. In the water regrid section, I include nst fields with the bundle if the nst flag is set, not in a separate bundle. I was able to cut out close to 1000 lines of code by using these bundles throughout the interp routine, even with the addition of the two subroutines I wrote to regrid and search over the bundles.

In total, I'm not sure creating a bundle with only nst fields and reusing this field bundle in two places would save that much code since there aren't any other places where *only* nst fields are looped over. Since the water regrid includes other fields, I'd have to create a bundle with only those non-nst fields and call the regrid and search routines twice in that case. 

Does any of this makes sense? If you want you can look at my other branch ([feature/sfc_regrid_many](https://github.com/LarissaReames-NOAA/UFS_UTILS/tree/feature/sfc_regrid_many), no PR yet) to see what my code looks like.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -4072,203 +3538,74 @@ subroutine nst_land_fill
...
- where(mask_ptr /= 0) data_ptr = 0.0
-
-! zm
-
- print*,"- CALL FieldGet FOR ZM."
- call ESMF_FieldGet(zm_target_grid, &
-                    farrayPtr=data_ptr, rc=rc)
- if(ESMF_logFoundError(rcToCheck=rc,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
-    call error_handler("IN FieldGet", rc)
+ do i = 1,16
```
GeorgeGayno-NOAA, line 3591: There is an ESMF routine that will return the number of fields (fieldCount) within a bundle:
http://earthsystemmodeling.org/docs/release/ESMF_8_1_1/ESMF_refdoc/node5.html#SECTION050251000000000000000

LarissaReames-NOAA: Yeah, I use that functionality in the new *****_many routines in the sfc_regrid_many branch, but in this case I know the field number exactly so I didn't really see the benefit to calling another ESMF routine in this case.

edwardhartnett: In that case please define a parameter NUM_ESMF_FIELDS or something of that sort. You should not use bare constants in production code (it's OK to do so in test code).

LarissaReames-NOAA: Done. I also converted the other bare constants that had been used in the code prior to my changes to parameters as well.

GeorgeGayno-NOAA:
Did you run the consistency tests? I can do so on the machines you can't access.

LarissaReames-NOAA:
@GeorgeGayno-NOAA They passed on Hera.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA They passed on Hera.

Merge the latest updates from 'develop'. Then I can run them today and, hopefully, merge.

LarissaReames-NOAA:
The branch has been updated and passes tests.

LarissaReames-NOAA:
So it looks like. while the consistency tests currently in the develop branch all pass, the tests I'd written for the regional capabilities (PR #252) are now broken. It appears this was caused by the implementation of the new tracers in PR #388 as these tests passed before I merged in the most recent develop branch changes

GeorgeGayno-NOAA:
> > @GeorgeGayno-NOAA They passed on Hera.
> 
> Merge the latest updates from 'develop'. Then I can run them today and, hopefully, merge.

Tests passed on Hera, Orion, Jet, WCOSS-Cray and WCOSS-Dell

LarissaReames-NOAA:
All develop branch and regional grib2 system/consistency tests now pass on Hera after implementing a few fixes for the new num_tracers_input usage. Unit tests also pass.

GeorgeGayno-NOAA:
> All develop branch and regional grib2 system/consistency tests now pass on Hera after implementing a few fixes for the new num_tracers_input usage. Unit tests also pass.

I had no problems running the consistency tests for this PR. Which tests failed for you?

LarissaReames-NOAA:
> > All develop branch and regional grib2 system/consistency tests now pass on Hera after implementing a few fixes for the new num_tracers_input usage. Unit tests also pass.
> 
> I had no problems running the consistency tests for this PR. Which tests failed for you?

The consistency tests that have been in a PR to the release branch for months and are just as suitable for the develop branch. They test a much wider range of capabilities that the code has for grib2 files, including reading in the Thompson aerosol variables. I've been using them to test as I develop for the develop branch. I'm going to submit a PR later today to the develop branch so we can get those in.

GeorgeGayno-NOAA:
@LarissaReames-NOAA Is this ready to merge?

LarissaReames-NOAA:
@GeorgeGayno-NOAA Yes, the tracer issues were fixed and I updated the test for reading the variable mapping table to check for the new num_tracers_input variable.

GeorgeGayno-NOAA (approved):

Issue #565, opened 2021-08-16
add relative humidity to specific humidity calculation for GFSv15/16 grib2


JiliDong-NOAA:
The current relative humidity to specific humidity conversion in chgres_cube (rh2spfh) is not consistent with GFSv15/16 grib2 files (when spfh is converted to rh by UPP in operations). A more consistent rh to spfh calculation could help to keep consistency when processing GFSv15/16 grib2 files (thanks for the discussions with Guoqing Ge of GSL, Fanglin Yang and Wen Meng of EMC). 

Furthermore, when there are missing levels of spfh (e.g. GFSV16 combined grib2), using rh with full levels could be a better option over using spfh with interpolation to missing levels.

A PR will be submitted shortly to add two namelist entries:
1. a parameter to choose if using rh or spfh in grib2 files for humidity ("use_rh" = false as default).
2. a parameter to choose which rh to spfh calculation will be used. ("calrh" = 0 as default to use the current calculation, =1 to use calculation consistent with GFSv15/16).

PR #566, opened 2021-08-16
add use_rh and calrh for grib2 humidity processing

JiliDong-NOAA:
This PR is to add two namelist entries and additional rh to spfh calculation for GFSv15/16 grib2:

"use_rh":  if using rh or spfh in grib2 files for humidity (False: use spfh if existing. True: use RH no matter spfh exists or not. Default is False)

"calrh":  which rh to spfh calculation will be used. ( = 0 as default to use the current calculation, =1 to use calculation consistent with GFSv15/16).

When setting these two parameters as default (or not specified in namelist), the regression tests won't change.

Related to #565


JiliDong-NOAA:
@GeorgeGayno-NOAA Not sure how to fix the Doxygen warning to pass check tests:

_/home/runner/work/UFS_UTILS/UFS_UTILS/ufs_utils/sorc/chgres_cube.fd/grib2_util.F90:147: error: argument 'fpvsnew' of command @param is not found in the argument list of grib2_util::fpvsnew(real, intent(in) t) (warning treated as error, aborting now)_

fpvsnew is a function to calculate saturation water vapor pressure. And I have:

_!> Compute saturation vapor pressure 
!!
!! @param[out] fpvsnew saturation vapor pressure in Pascals 
!! @param[in] t temperature in Kelvin
!! @author N Phillips_

JiliDong-NOAA:
> @GeorgeGayno-NOAA Not sure how to fix the Doxygen warning to pass check tests:
> 
> _/home/runner/work/UFS_UTILS/UFS_UTILS/ufs_utils/sorc/chgres_cube.fd/grib2_util.F90:147: error: argument 'fpvsnew' of command @param is not found in the argument list of grib2_util::fpvsnew(real, intent(in) t) (warning treated as error, aborting now)_
> 
> fpvsnew is a function to calculate saturation water vapor pressure. And I have:
> 
> _!> Compute saturation vapor pressure !! !! @param[out] fpvsnew saturation vapor pressure in Pascals !! @param[in] t temperature in Kelvin !! @author N Phillips_

When removing !! @param[out] fpvsnew, I got the following Doxygen error:

/home/runner/work/UFS_UTILS/UFS_UTILS/ufs_utils/sorc/chgres_cube.fd/grib2_util.F90:154: error: return type of member grib2_util::fpvsnew is not documented (warning treated as error, aborting now)

GeorgeGayno-NOAA:
> > @GeorgeGayno-NOAA Not sure how to fix the Doxygen warning to pass check tests:
> > _/home/runner/work/UFS_UTILS/UFS_UTILS/ufs_utils/sorc/chgres_cube.fd/grib2_util.F90:147: error: argument 'fpvsnew' of command @param is not found in the argument list of grib2_util::fpvsnew(real, intent(in) t) (warning treated as error, aborting now)_
> > fpvsnew is a function to calculate saturation water vapor pressure. And I have:
> > _!> Compute saturation vapor pressure !! !! @param[out] fpvsnew saturation vapor pressure in Pascals !! @param[in] t temperature in Kelvin !! @author N Phillips_
> 
> When removing _!! @param[out] fpvsnew_, I got the following Doxygen error:
> 
> _/home/runner/work/UFS_UTILS/UFS_UTILS/ufs_utils/sorc/chgres_cube.fd/grib2_util.F90:154: error: return type of member grib2_util::fpvsnew is not documented (warning treated as error, aborting now)_

I will take a look.

GeorgeGayno-NOAA:
@JiliDong-NOAA For functions, doxygen requires a "@return" entry. Here is what I did to fix it:

```
--- a/sorc/chgres_cube.fd/grib2_util.F90
+++ b/sorc/chgres_cube.fd/grib2_util.F90
@@ -146,6 +146,7 @@ end subroutine RH2SPFH_GFS
 !> Compute saturation vapor pressure
 !!
 !! @param[in] t temperature in Kelvin
+!! @return fpvsnew Saturation vapor pressure
 !! @author N Phillips

```

JiliDong-NOAA:
> @JiliDong-NOAA For functions, doxygen requires a "@return" entry. Here is what I did to fix it:
> 
> ```
> --- a/sorc/chgres_cube.fd/grib2_util.F90
> +++ b/sorc/chgres_cube.fd/grib2_util.F90
> @@ -146,6 +146,7 @@ end subroutine RH2SPFH_GFS
>  !> Compute saturation vapor pressure
>  !!
>  !! @param[in] t temperature in Kelvin
> +!! @return fpvsnew Saturation vapor pressure
>  !! @author N Phillips
> ```

Thanks @GeorgeGayno-NOAA . Now the check tests all passed. BTW, is there a way to do the compiling with debug options including Doxygen on Linux? Having to debug it online is sort of frustrating.

GeorgeGayno-NOAA:
> > @JiliDong-NOAA For functions, doxygen requires a "@return" entry. Here is what I did to fix it:
> > ```
> > --- a/sorc/chgres_cube.fd/grib2_util.F90
> > +++ b/sorc/chgres_cube.fd/grib2_util.F90
> > @@ -146,6 +146,7 @@ end subroutine RH2SPFH_GFS
> >  !> Compute saturation vapor pressure
> >  !!
> >  !! @param[in] t temperature in Kelvin
> > +!! @return fpvsnew Saturation vapor pressure
> >  !! @author N Phillips
> > ```
> 
> Thanks @GeorgeGayno-NOAA . Now the check tests all passed. BTW, is there a way to do the compiling with debug options including Doxygen on Linux? Having to debug it online is sort of frustrating.

In the 'build_all.sh' script, turn on the following CMAKE_FLAG: `-DENABLE_DOCS=ON`

JiliDong-NOAA:
@GeorgeGayno-NOAA @LarissaReames-NOAA @JeffBeck-NOAA could you review this PR? Thanks

GeorgeGayno-NOAA (changes requested):
@JiliDong-NOAA We are now committed to doing unit tests. So your new functions will need some tests.

sorc/chgres_cube.fd/grib2_util.F90
```diff
@@ -57,6 +57,202 @@ subroutine rh2spfh(rh_sphum,p,t)
...
+
+!
+!-------------------------------------------------------------------------------------
+!
+      elemental function fpvsnew(t)
+!$$$     Subprogram Documentation Block
+!
+! Subprogram: fpvsnew         Compute saturation vapor pressure
+!   Author: N Phillips            w/NMC2X2   Date: 30 dec 82
+!
```
GeorgeGayno-NOAA: We are trying to get rid of these old prologs. If any of this information is worth keeping, add it to the doxygen.

sorc/chgres_cube.fd/program_setup.F90
```diff
@@ -88,13 +88,15 @@ module program_setup
  integer, public                 :: cycle_hour = -999 !< Cycle hour.
  integer, public                 :: regional = 0 !<  For regional target grids.  When '1' remove boundary halo region from atmospheric/surface data and
                                                  !! output atmospheric boundary file. When '2' output boundary file only. Default is '0' (global grids).
+ integer, public                 :: calrh = 0 !< rh to spfh calculation. 0:original; 1: GFSv15/16
```
GeorgeGayno-NOAA: Instead of setting two new flags - calrh and use_rh - can this option be communicated thru the varmap table somehow?

JiliDong-NOAA: I don't know how to do that. These two parameters are only for humidity and not for other tracers/variables. Not sure if we could add extra columns only for one particular variable in varmap table. I wonder if @LarissaReames-NOAA and @JeffBeck-NOAA have any thoughts on this.

JiliDong-NOAA:
> @JiliDong-NOAA We are now committed to doing unit tests. So your new functions will need some tests.

@GeorgeGayno-NOAA Unit test (ftst_rh2spfh_gfs) added and passed for new rh to spfh subroutines.

JeffBeck-NOAA:
@JiliDong-NOAA, based on your issue for this PR, I assume FV3 can initialize off of RH *or* SPFH?  Also, can you comment on how the new RH-to-SPFH calculation differs from what's already in chgres_cube, and whether that would be best for all external model data?  Thanks!

sorc/chgres_cube.fd/program_setup.F90
JeffBeck-NOAA: After discussing with @LarissaReames-NOAA, we believe that "calrh" and "use_rh" should stay as namelist entries, since the varmap table is really for handling of fields that are required by the CCPP scheme the user wishes to run, while this is a variable calculation option, which depends on how someone wishes to initialize.

LarissaReames: What if we make these namelist options for super users who wish to have complete control, but default to what we think is the "best practice" options based on data type, external model, and data date? So in this case, switch to the new calculation type for grib2, GFS, post-whatever date the new GFS data was implemented in the grib2 files. This would keep the vast majority of users from ever having to worry about specifying these options.

JiliDong-NOAA:
> @JiliDong-NOAA, based on your issue for this PR, I assume FV3 can initialize off of RH _or_ SPFH? Also, can you comment on how the new RH-to-SPFH calculation differs from what's already in chgres_cube, and whether that would be best for all external model data? Thanks!

@JeffBeck-NOAA I will upload several slides to show the difference.

JiliDong-NOAA:
> > @JiliDong-NOAA, based on your issue for this PR, I assume FV3 can initialize off of RH _or_ SPFH? Also, can you comment on how the new RH-to-SPFH calculation differs from what's already in chgres_cube, and whether that would be best for all external model data? Thanks!
> 
> @JeffBeck-NOAA I will upload several slides to show the difference.

@JeffBeck-NOAA Here are the comparison of different rh to spfh conversions, verified against original spfh as truth by calculating rms difference between them for three sets of grib2 files (GFSv15, v16 and grib2 from rt). Also compared with interpolating 41 levels spfh to 57 levels for one set grib2 file. Using the newly added rh2spfh conversion shows the smallest rms differences. 

The new conversion is directly extracted from UPP/EMC_post and consistent with operational GFSv15/16.  

Some recommendations are listed in the last slide.

[qv_for GFS_grib2.pdf](https://github.com/NOAA-EMC/UFS_UTILS/files/7024096/qv_for.GFS_grib2.pdf)

sorc/chgres_cube.fd/program_setup.F90
JiliDong-NOAA: @LarissaReames-NOAA good idea. But I am not sure which option fits most scenarios. I have some recommendations as follows:

For grib2 full level qv, use qv (e.g. GFSv15 combined, GFSv16 uncombined) 
For grib2 qv with missing levels or not available, use rh with consistent conversion (e.g. GFSv15 uncombined, GFSv16 combined)

As for specific GFS grib2 data: 
GFSv15 uncombined: use_rh=True; calrh=1 
GFSv15 combined: use_rh=False 
GFSv16 uncombined: use_rh=False 
GFSv16 combined: use_rh=True; calrh=1

Maybe we can set default as "use_rh=True; calrh=1" since this may be the best option for the current operational GFS grib2 (combined).

JiliDong-NOAA: Made a couple of changes:
1. remove calrh and use external_model to determine rh to spfh conversion: if external_model is GFS, rh2spfh_gfs will be used;
2. default use_rh as True as the option for current combined operational GFSV16 grib2 files

JiliDong-NOAA: The changes will need the following regression test baseline update since now the default set up uses RH instead of SPFH for grib2 files:

consistency.log08:<<< C192 GFS GRIB2 TEST FAILED. >>>
consistency.log09:<<< 25-KM CONUS GFS GRIB2 TEST FAILED. >>>
consistency.log10:<<< 3-km CONUS HRRR W/ GFS PHYSICS GRIB2 TEST FAILED. <<<
consistency.log11:<<< 3-km CONUS HRRR W/ GSD PHYSICS AND SFC FROM FILE GRIB2 TEST FAILED. <<<
consistency.log13:<<< 13-km CONUS RAP W/ GSD PHYSICS AND SFC FROM FILE GRIB2 TEST FAILED. <<<
consistency.log14:<<< 13-KM NA GFS NCEI GRIB2 TEST FAILED. >>>
consistency.log16:<<< 25-KM CONUS GFS PGRIB2+BGRIB2 TEST FAILED. >>>

Please let me know how you would like me to proceed.

LarissaReames: > Made a couple of changes:
> 
> 1. remove calrh and use external_model to determine rh to spfh conversion: if external_model is GFS, rh2spfh_gfs will be used;
> 2. default use_rh as True as the option for current combined operational GFSV16 grib2 files

1: Is this really the correct calculation for *all* GFS data, or just for dates after the switch to V15?

2: So the majority of files that chgres_cube is capable of reading do actually have spfh, so I think it makes much more sense to have use of spfh be the default.

As for your failed tests, I certainly feel like if the files have all levels of spfh already, and not rh, but you've chosen to use RH, then the code should quit and provide an error. The functionalities that you added should not be breaking RAP and HRRR consistency tests since your code is not relevant to these files at all.

JeffBeck-NOAA: @JiliDong-NOAA, since all of this is specific to the GFS, I think we should only be modifying "RH" vs "SPFH" for GFS grib2 data based on date, therefore RAP- and HRRR-based grib2 file processing should stay the same.

Also, there seems to be a discrepancy here:

As for specific GFS grib2 data:
GFSv15 uncombined: use_rh=True; calrh=1
GFSv15 combined: use_rh=False
GFSv16 uncombined: use_rh=False
GFSv16 combined: use_rh=True; calrh=1

Use of "RH" is flipped for uncombined vs. combined from v15 to v16?

JeffBeck-NOAA:
> > > @JiliDong-NOAA, based on your issue for this PR, I assume FV3 can initialize off of RH _or_ SPFH? Also, can you comment on how the new RH-to-SPFH calculation differs from what's already in chgres_cube, and whether that would be best for all external model data? Thanks!
> > 
> > 
> > @JeffBeck-NOAA I will upload several slides to show the difference.
> 
> @JeffBeck-NOAA Here are the comparison of different rh to spfh conversions, verified against original spfh as truth by calculating rms difference between them for three sets of grib2 files (GFSv15, v16 and grib2 from rt). Also compared with interpolating 41 levels spfh to 57 levels for one set grib2 file. Using the newly added rh2spfh conversion shows the smallest rms differences.
> 
> The new conversion is directly extracted from UPP/EMC_post and consistent with operational GFSv15/16.
> 
> Some recommendations are listed in the last slide.
> 
> [qv_for GFS_grib2.pdf](https://github.com/NOAA-EMC/UFS_UTILS/files/7024096/qv_for.GFS_grib2.pdf)

Thanks, @JiliDong-NOAA.  It certainly looks like the new formulation of SPFH is much better!

sorc/chgres_cube.fd/program_setup.F90
JiliDong-NOAA: I made another change so that using rh or spfh in grib2 will be based on if spfh are on the same level as UGRD: if not, rh will be used. "use_rh" is also removed. The current code doesn't have any additional parameters. The regression test baseline needs to be updated for GFS grib2:

consistency.log08:<<< C192 GFS GRIB2 TEST FAILED. >>>
consistency.log09:<<< 25-KM CONUS GFS GRIB2 TEST FAILED. >>>
consistency.log14:<<< 13-KM NA GFS NCEI GRIB2 TEST FAILED. >>>
consistency.log16:<<< 25-KM CONUS GFS PGRIB2+BGRIB2 TEST FAILED. >>>

JiliDong-NOAA: > > Made a couple of changes:
> > 
> > 1. remove calrh and use external_model to determine rh to spfh conversion: if external_model is GFS, rh2spfh_gfs will be used;
> > 2. default use_rh as True as the option for current combined operational GFSV16 grib2 files
> 
> 1: Is this really the correct calculation for _all_ GFS data, or just for dates after the switch to V15?
> 
> 2: So the majority of files that chgres_cube is capable of reading do actually have spfh, so I think it makes much more sense to have use of spfh be the default.
> 
> As for your failed tests, I certainly feel like if the files have all levels of spfh already, and not rh, but you've chosen to use RH, then the code should quit and provide an error. The functionalities that you added should not be breaking RAP and HRRR consistency tests since your code is not relevant to these files at all.

rh2spfh_gfs is for GFSv15/16 or FV3GFS.

JiliDong-NOAA: > @JiliDong-NOAA, since all of this is specific to the GFS, I think we should only be modifying "RH" vs "SPFH" for GFS grib2 data based on date, therefore RAP- and HRRR-based grib2 file processing should stay the same.
> 
> Also, there seems to be a discrepancy here:
> 
> As for specific GFS grib2 data:
> GFSv15 uncombined: use_rh=True; calrh=1
> GFSv15 combined: use_rh=False
> GFSv16 uncombined: use_rh=False
> GFSv16 combined: use_rh=True; calrh=1
> 
> Use of "RH" is flipped for uncombined vs. combined from v15 to v16?

After the most recent change, HRRR/RAP processing should stay the same if SPFH are on the same levels as UGRD. For the last question, it is not flipped. GFS v15 uncombined grib2 has no spfh so rh is used. GFSv16 combined has missing levels of spfh and rh is preferred. Both GFSv15 combined and GFSv16 uncombined has spfh on the same levels as UGRD and are considered as full levels spfh.

JiliDong-NOAA: I just confirmed with Fanglin that all GFS versions use the same conversion of spfh to rh.

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -2603,11 +2605,22 @@ subroutine read_input_atm_grib2_file(localpet)
      if (localpet==0) print*, "- LEVEL AFTER SORT = ",slevs(i)
    enddo
  endif
+
+! Is SPFH on full levels Jili Dong
+     do vlev = 1, lev_input
+      iret = grb2_inq(the_file,inv_file,':SPFH:',slevs(vlev))
+      if (iret <= 0) then
+        use_rh = .TRUE.
+        if (localpet == 0) print*,':SPFH:'//slevs(vlev)//' not existing. Use RH instead.'
```
LarissaReames: Let's update the notification message to be: `':SPFH on level '//trim(slevs(vlev))//' does not exist. Will read in RH and convert to SPFH instead.'` This makes the message shorter (without trim() you get a bunch of empty space) and makes it clear that RH will be read in and later converted to SPFH, not actually used in the output data. I was sort of confused when I first read the output messages, so this change in description might be helpful

JiliDong-NOAA: updated as suggested. Thanks @LarissaReames-NOAA

sorc/chgres_cube.fd/program_setup.F90
JiliDong-NOAA: The current code update allows chgres to:

- automatically choose reading SPFH or RH based on whether grib2 data contains SPFH at full levels as UGRD
- automatically choose which rh to spfh conversions based on external_model (GFS vs. non-GFS)

The previous added parameters (use_rh and calrh) are removed.

JiliDong-NOAA:
@GeorgeGayno-NOAA could you update the new chgres rt baseline related to GFS grib2? I have the new runs on Hera:

/scratch2/NCEPDEV/stmp1/Jili.Dong/reg-tests/chgres-cube/
c192_gfs_grib2 (Test 8)
25km_conus_gfs_grib2 (Test 9)
13km_na_gfs_ncei_grib2 (Test 14)
25km_conus_gfs_pbgrib2 (Test 16)

Thanks.

JiliDong-NOAA:
@GeorgeGayno-NOAA Please let me know if there is any change you would like me to make.

GeorgeGayno-NOAA:
@JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.

JiliDong-NOAA:
> @JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.

No, rh2spfh_nam is not used for now. It is for the future GFSv17.

JeffBeck-NOAA:
> > @JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.
> 
> No, rh2spfh_nam is not used for now. It is for the future GFSv17.

@JiliDong-NOAA, why use another formulation for GFSv17?  Is there another way to calculate spfh from rh that is better than what is proposed in this PR?

JiliDong-NOAA:
> > > @JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.
> > 
> > 
> > No, rh2spfh_nam is not used for now. It is for the future GFSv17.
> 
> @JiliDong-NOAA, why use another formulation for GFSv17? Is there another way to calculate spfh from rh that is better than what is proposed in this PR?

GFSv17 will use a different spfh to rh conversion from the previous GFS versions. For chgres, I think we would like to apply a conversion consistent with the parent or original model to be ingested (either GFS or HRRR) when calculating spfh back from rh. In this context, I think a "consistent" conversion means a "better" solution here.

GeorgeGayno-NOAA:
> > > > @JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.
> > > 
> > > 
> > > No, rh2spfh_nam is not used for now. It is for the future GFSv17.
> > 
> > 
> > @JiliDong-NOAA, why use another formulation for GFSv17? Is there another way to calculate spfh from rh that is better than what is proposed in this PR?
> 
> GFSv17 will use a different spfh to rh conversion from the previous GFS versions. For chgres, I think we would like to apply a conversion consistent with the parent or original model to be ingested (either GFS or HRRR) when calculating spfh back from rh. In this context, I think a "consistent" conversion means a "better" solution here.

@JiliDong-NOAA GFS v17 parallels may not start for a while. And there is no logic to detect GFS v17 data and call that function accordingly. So, until then it a 'dead' routine. If you insist on keeping it as part of chgres, I would at least rename it - something like `rh2spfh_gfsv17`.  Since there are multiple rh to spfh routines now, I would give each one a descriptive name. And provide enough details in the prolog so users know the difference between each function.

JiliDong-NOAA:
> > > > > @JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.
> > > > 
> > > > 
> > > > No, rh2spfh_nam is not used for now. It is for the future GFSv17.
> > > 
> > > 
> > > @JiliDong-NOAA, why use another formulation for GFSv17? Is there another way to calculate spfh from rh that is better than what is proposed in this PR?
> > 
> > 
> > GFSv17 will use a different spfh to rh conversion from the previous GFS versions. For chgres, I think we would like to apply a conversion consistent with the parent or original model to be ingested (either GFS or HRRR) when calculating spfh back from rh. In this context, I think a "consistent" conversion means a "better" solution here.
> 
> @JiliDong-NOAA GFS v17 parallels may not start for a while. And there is no logic to detect GFS v17 data and call that function accordingly. So, until then it a 'dead' routine. If you insist on keeping it as part of chgres, I would at least rename it - something like `rh2spfh_gfsv17`. Since there are multiple rh to spfh routines now, I would give each one a descriptive name. And provide enough details in the prolog so users know the difference between each function.

@GeorgeGayno-NOAA Sure, I will have rh2gfs_nam removed. Good suggestion to add some comments/details on those subroutines. I am working on it.

JiliDong-NOAA:
> > > > > > @JiliDong-NOAA Is `rh2spfh_nam` used? I don't see it referenced anywhere.
> > > > > 
> > > > > 
> > > > > No, rh2spfh_nam is not used for now. It is for the future GFSv17.
> > > > 
> > > > 
> > > > @JiliDong-NOAA, why use another formulation for GFSv17? Is there another way to calculate spfh from rh that is better than what is proposed in this PR?
> > > 
> > > 
> > > GFSv17 will use a different spfh to rh conversion from the previous GFS versions. For chgres, I think we would like to apply a conversion consistent with the parent or original model to be ingested (either GFS or HRRR) when calculating spfh back from rh. In this context, I think a "consistent" conversion means a "better" solution here.
> > 
> > 
> > @JiliDong-NOAA GFS v17 parallels may not start for a while. And there is no logic to detect GFS v17 data and call that function accordingly. So, until then it a 'dead' routine. If you insist on keeping it as part of chgres, I would at least rename it - something like `rh2spfh_gfsv17`. Since there are multiple rh to spfh routines now, I would give each one a descriptive name. And provide enough details in the prolog so users know the difference between each function.
> 
> @GeorgeGayno-NOAA Sure, I will have rh2gfs_nam removed. Good suggestion to add some comments/details on those subroutines. I am working on it.

@GeorgeGayno-NOAA rh2spfh_nam removed and documentation added for different rh2spfh conversions. Thanks.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA could you update the new chgres rt baseline related to GFS grib2? I have the new runs on Hera:
> 
> /scratch2/NCEPDEV/stmp1/Jili.Dong/reg-tests/chgres-cube/
> c192_gfs_grib2 (Test 8)
> 25km_conus_gfs_grib2 (Test 9)
> 13km_na_gfs_ncei_grib2 (Test 14)
> 25km_conus_gfs_pbgrib2 (Test 16)
> 
> Thanks.

Thanks. I will update the baseline directories.

GeorgeGayno-NOAA (approved):
I have no further comments. @JeffBeck-NOAA and @LarissaReames-NOAA are the experts with the grib2 option. I will merge after they also approve.

JiliDong-NOAA:
@GeorgeGayno-NOAA @JeffBeck-NOAA @LarissaReames-NOAA Please let me know if there is anything I can do for this PR.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA @JeffBeck-NOAA @LarissaReames-NOAA Please let me know if there is anything I can do for this PR.

It looks good to me. But since I am not an expert on the grib2 option, I would like either @JeffBeck-NOAA and/or @LarissaReames-NOAA to give their approval.

LarissaReames (approved):

GeorgeGayno-NOAA:
Thanks, @LarissaReames-NOAA 

@JiliDong-NOAA I will merge.

JiliDong-NOAA:
> Thanks, @LarissaReames-NOAA
> 
> @JiliDong-NOAA I will merge.

Thanks for you all.

Issue #591, opened 2021-10-21
Can we remove wgrib2 from the dependencies?


edwardhartnett:
Somehow wgrib2 crept into our dependencies.

We are trying to eliminate use of wgrib2 in NCEPLIBS. The wgrib2 library is not well-maintained, distributed, or tested, and there is no interest on the wgrib2 team in meeting our needs in these areas. So we need to not build more tools on top of wgrib2.

kgerheiser:
UFS_UTILS definitely depends on the wgrib2 library. That hasn't changed recently. The alternative is to port UFS_UTILS to use g2 as its grib2 library.

GeorgeGayno-NOAA:
chgres_cube uses wgrib2. It could be modified to use the 'g2' library. I don't think that would be difficult. But it would take some work and testing. And it could be done is smaller chunks.

edwardhartnett:
@GeorgeGayno-NOAA for future work, can you communicate to the other chgres_cube programmers that wgrib2 should be avoided?

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA for future work, can you communicate to the other chgres_cube programmers that wgrib2 should be avoided?

Absolutely.

Maybe I can add a warning to our wiki page?  https://github.com/NOAA-EMC/UFS_UTILS/wiki
Or the PR template? https://github.com/NOAA-EMC/UFS_UTILS/blob/develop/.github/PULL_REQUEST_TEMPLATE

edwardhartnett:
If we can remove the dependency on wgrib2, we should.

GeorgeGayno-NOAA:
I added a warning page to the wiki:
https://github.com/ufs-community/UFS_UTILS/wiki/A-Warning-about-the-WGRIB2-library

XianwuXue-NOAA:
@GeorgeGayno-NOAA Based on the warning page "Pull requests that use WGRIB2 will be rejected.", do it also apply to the GEFS WCOSS2 Transition? Because GEFS uses chgres_cube.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA Based on the warning page "Pull requests that use WGRIB2 will be rejected.", do it also apply to the GEFS WCOSS2 Transition? Because GEFS uses chgres_cube.

Are you adding new code to chgres_cube that uses wgrib2?

XianwuXue-NOAA:
[...]

PR #625, opened 2022-02-02
chgres_cube - remove wgrib2 from the GRIB2 data read routines.

GeorgeGayno-NOAA:
## DESCRIPTION OF CHANGES: 

Begin removal of WGRIB2 from chgres_cube:
- Update routine `read_input_atm_grib2_file`, `read_input_sfc_grib2_file`,  `read_winds` and `read_grib_soil` to read input GRIB2 files using G2LIB instead of WGRIB2.
- Update routine `define_input_grid_gfs_grib2` to flip the pole of GFS GRIB2 data. WGRIB2 and G2LIB use different conventions for defining global gaussian grids.
- Update to G2LIB v3.4.3 or higher, which eliminated occasional segmentation faults and slow wall clock times.
- Two new units test for GRIB2 read routines `read_input_atm_grib2_file` and `read_input_sfc_grib2_file`.

## TESTS CONDUCTED: 
Two new unit tests were created. 

All consistency tests were run on all supported machines. The GRIB2 tests failed as expected. But differences between the baseline files and the test files were small. Will provide location of all log files shortly.

## DEPENDENCIES:
None.

## DOCUMENTATION:
N/A

## ISSUE: 
Part of #591. This is too big a project for one PR. Additional PRs under that issue will follow.


kgerheiser:
This will simplify library maintenance tremendously. Installing and maintaining wgrb2 as a library (and executable) was a real pain.

GeorgeGayno-NOAA:
Logs from the failed consistency tests are on Hera: /scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/wgrib2_logs

I can provide the output files and their differences from the baseline for visualizing if you wish.

GeorgeGayno-NOAA:
> Logs from the failed consistency tests are on Hera: /scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/wgrib2_logs
> 
> I can provide the output files and their differences from the baseline for visualizing if you wish.

I wrote some scripts to create some difference plots. All plots should be done by next Tuesday.

GeorgeGayno-NOAA:
All test output is on Hera - /scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/chgres_cube_nowgrib2

There is a subdirectory for each failed GRIB2 consistency test. The directory contains the output log file from the test, which lists all fields that differ from the baseline and a summary of those differences (via the `nccmp` utility).  Under the ./graphics subdirectory there are some Grads plots of difference fields.

@JeffBeck-NOAA and @LarissaReames-NOAA - please take a look. Let me know if you want to see any other plots. And recall, this PR is only for the conversion of the read_data module and not the model_grid module. So there are no lat/lon difference plots.

sorc/chgres_cube.fd/input_data.F90
```diff
@@ -4842,665 +5211,925 @@ subroutine read_input_sfc_grib2_file(localpet)
...
    endif
 
+   if (.not. minmax_vgfrc_from_climo) then
+
+     if (localpet == 0) then
+
+       print*,"- READ MIN VEG FRACTION."
+
+       jdisc   = 2  ! Search for discipline - land products
+       j = 1105 ! grib2 file does not distinguish between the various veg
```
LarissaReames, line 5660: I don't know if we want to address this in the current PR, or open a new one after this one is approved, but in the more recent HRRR files they actually have upgraded to using the correct wgrib2 parameter numbers for min and max veg fraction (231 and 232, respectively). Do we want to put that in as the first try for these variables here or wait and put it in a new PR? It wouldn't change any of the baseline results as the current HRRR file being used for baseline testing is old so it still needs these record number specifications. I'm honestly not sure how we haven't had an Issue submitted due to failure caused by the upgrade to the newer (and better) parameter numbers.

GeorgeGayno-NOAA: I might do this in a separate PR. 

Ideally, I would like all these static fields created at the grid generation step (via sfc_climo_gen). Do we have the original greenness dataset used by HRRR? Converting it to the format used by sfc_climo_gen is not difficult.

JeffBeck-NOAA (approved):
Difference fields between wgrib2 and g2 results look reasonable (only noise) and on par with what is expected for nearly identical input grid point locations.

GeorgeGayno-NOAA:
Thanks, @JeffBeck-NOAA. I will merge now.

PR #692, opened 2022-09-06
Increase pull data timeout

### LINKED ISSUE #691: Cmake tests timeout while downloading data
Cmake tests often fail due to the test data timing out before the download is complete. The timeout in the pull_data function should be increased so the download can complete before being killed.

WalterKolczynski-NOAA:
## DESCRIPTION OF CHANGES: 
Tests would often fail because the data would not finish transferring before the timeout threshold was reached. That timeout is now increased so data should finish transferring in time.

## TESTS CONDUCTED: 
- [x] cmake tests all pass on Hera for Intel

## DEPENDENCIES:
None

## DOCUMENTATION:
None

## ISSUE: 
Fixes #691

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -14,7 +14,7 @@ function(PULL_DATA THE_URL THE_FILE)
       ${CMAKE_CURRENT_BINARY_DIR}/data/${THE_FILE}
       SHOW_PROGRESS
       STATUS status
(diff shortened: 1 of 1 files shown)


JeffBeck-NOAA (approved):

GeorgeGayno-NOAA:
Thanks @WalterKolczynski-NOAA. Will merge.

PR #704, opened 2022-10-18
Modify chgres_cube output netcdf4 files.

### LINKED ISSUE #689: The new function of chgres_cube to output the results in netcdf 4 format
Currently, the chgres_cube is generating results in netcdf 4 classic with hardwired setup. 
This issue is opened for adding the function to generate netcdf 4 files.  There are two reasons for this request: 1. The chgres_cube generated FV3-LAM cold start files would be read in and updated by GSI. Currently the GSI IO interface for fv3-lam (including the cold start files generated by chgres_cube ) is using the parallel IO of netcdf 4.  When the netcdf 4 classic files are used for GSI,  some minor differences would generated in the final analysis fields.
Hence, it is expected chgres_cube can  create netcdf 4 files.  It would be very helpful if this new function with chgres_cube could take care of the chunk size /shape in the generated netcdf 4 files, because this has an significant impact on the performance of GSI parallel IO for fv3-lam and it could also impact the IO for FV3 model. 
2. For the fv3-lam forecast model , It was found the cold start files in netcdf 4 class would cause slower initialization part compared with runs using restart files in netcdf 4 (though the reasons are still to be identified).  
Hence , that will be very helpful if chgres_cube would be implemented with the function to generate files in netcdf 4 format . 
I also cc our colleagues directly involved in this issue  Shun Liu  , Ming Hu (@hu5970 ) and Eric Rodger . And they can clarify on this problem.

GeorgeGayno-NOAA:
## DESCRIPTION OF CHANGES: 
Modify `chgres_cube` to output files in netcdf4 (and not netcdf4-classic).

## TESTS CONDUCTED: 
The consistency tests were run on Hera, Jet, Orion and WCOSS2. All tests failed (according to the `nccmp` utility) because the baseline files were netcdf4-classic. That was the only difference. All data values were the same. This result was expected.

## DEPENDENCIES:
None.

## DOCUMENTATION:
N/A

## ISSUE: 
Fixes #689.

### DIFF
--- a/sorc/chgres_cube.fd/write_data.F90
+++ b/sorc/chgres_cube.fd/write_data.F90
@@ -41,7 +41,6 @@
 
  character(len=13)   :: outfile
 
- integer             :: fsize=65536, initial = 0
  integer             :: header_buffer_val = 16384
  integer             :: error, ncid, dim_nvcoord
  integer             :: dim_levp1, id_ntrac, id_vcoord
@@ -55,8 +54,7 @@
 
  print*,"- WRITE ATMOSPHERIC HEADER FILE: ", trim(outfile)
 
- error = nf90_create(outfile, IOR(NF90_NETCDF4,NF90_CLASSIC_MODEL), &
-                     ncid, initialsize=initial, chunksize=fsize)
+ error = nf90_create(outfile, NF90_NETCDF4, ncid)
  call netcdf_err(error, 'CREATING FILE='//trim(outfile) )
 
  error = nf90_def_dim(ncid, 'nvcoord', nvcoord_target, dim_nvcoord)
@@ -135,7 +133,6 @@
 
  character(len=50)              :: name
 
- integer                        :: fsize=65536, initial = 0
  integer                        :: header_buffer_val = 16384
  integer                        :: ncid, error, tile, i, n
  integer                        :: dim_lon, dim_lat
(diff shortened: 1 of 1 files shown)


edwardhartnett (approved):

GeorgeGayno-NOAA:
Will merge, then update the consistency test baseline files on all machines.

sorc/chgres_cube.fd/write_data.F90
```diff
@@ -55,8 +54,7 @@ subroutine write_fv3_atm_header_netcdf(localpet)
 
  print*,"- WRITE ATMOSPHERIC HEADER FILE: ", trim(outfile)
 
- error = nf90_create(outfile, IOR(NF90_NETCDF4,NF90_CLASSIC_MODEL), &
-                     ncid, initialsize=initial, chunksize=fsize)
+ error = nf90_create(outfile, NF90_NETCDF4, ncid)
```
edwardhartnett, line 57: This change loses the NF90_CLASSIC_MODEL flag, and that's fine.

The CLASSIC_MODEL flag ensures that the created file follows netCDF classic rules, so it can be processed by netcdf-classic (i.e. pre-netCDF4) code. Although the CLASSIC_MODEL enforces some restrictions on the netCDF-4 file, it is not really necessary.

If we refrain from using netCDF-4 user-defined types, hierarchical groups, and unsigned integer types (and this is the current plan, AFAIK), then the output files will continue to be readable by netCDF classic code. The NC4_CLASSIC_MODEL flag is not needed for this, we will still get it.

Eliminating the initialsize and chunksize optional parameters is good. They are ignored for NC_NETCDF4 files. NetCDF-4 chunksizes are set on a per-variable basis, when the variable is defined. If you don't explicitly set the variable chunksizes, then netCDF will take a guess and come up with chunksizes for each var.

So this is a good change.

Issue #705, opened 2022-10-18
Incorporate BNU soil type data


GeorgeGayno-NOAA:
Incorporate the BNU soil type data to support the regional model.

Data is available here: https://www2.mmm.ucar.edu/wrf/users/download/get_sources_wps_geog.html#optional

GeorgeGayno-NOAA:
Data was pulled to Cactus using this command:

`wget https://www2.mmm.ucar.edu/wrf/src/wps_files/geog_alt_lsm.tar.gz`

The tar file was decompressed, then untarred:

`gzip -d geog_alt_lsm.tar.gz`
`tar -xvf geog_alt_lsm.tar`

All data was placed in `/lfs/h2/emc/global/noscrub/George.Gayno/ufs_utils.git/bnu.soil/orig_data`

GeorgeGayno-NOAA:
Rongqian has created a file from the original dataset, not from the WRF tiles as I did. Hold further work for now.

barlage:
@GeorgeGayno-NOAA The global 30" BNU soil type data in the same format as the global 30" VIIRS vegetation are stored in the land data space on hera: /scratch2/NCEPDEV/land/data/input_data/soil_type/soil_type.bnu.30s.nc

Note this is not internally compressed. If you want me to compress George, let me know. I will follow up with some more documentation.

barlage:
@GeorgeGayno-NOAA for documentation and scripts, do you want a new directory added to the util directory with a readme and scripts?

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA The global 30" BNU soil type data in the same format as the global 30" VIIRS vegetation are stored in the land data space on hera: /scratch2/NCEPDEV/land/data/input_data/soil_type/soil_type.bnu.30s.nc
> 
> Note this is not internally compressed. If you want me to compress George, let me know. I will follow up with some more documentation.

Thanks, I will run a test grid with this data.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA for documentation and scripts, do you want a new directory added to the util directory with a readme and scripts?

We will add a new sub-directory under 'util': https://github.com/ufs-community/UFS_UTILS/tree/develop/util

It can contain a readme file and the source code. I would also include a simple build script and run script for the machine you used to create the data. How much source code/scripts are associated with this project?

barlage:
[...]

PR #707, opened 2022-10-26
Add to develop branch readme, scripts and code to create bnu soil texture data

barlage:
This PR adds documentation for how to create BNU soil texture class data from the original sand/silt/clay percentages. 

## DESCRIPTION OF CHANGES: 
Add a directory to utils/ that documents the process of creating the BNU soil texture class data. No changes to the main UFS_UTILS code are made.

## TESTS CONDUCTED: 

- @RongqianYang-NOAA tested final product in UFS_UTILS on C96, C384 and C768 grids

## DEPENDENCIES:

None

## DOCUMENTATION:

Documentation specific to the utility added in util/bnu_soil_gen

## ISSUE: 

Related issue #705

### DIFF
--- a/util/bnu_soil_gen/README
+++ b/util/bnu_soil_gen/README
@@ -1,0 +1,144 @@
+
+Steps to create soil texture class dataset from BNU sand/silt/clay data source
+
+All processing scripts and code are run on NOAA hera HPC
+
+=======================================
+
+1. download original sand,silt,clay data from here:
+	http://globalchange.bnu.edu.cn/research/soilw
+
+6 files needed:
+
+CLAY1.nc  SAND1.nc  SILT1.nc
+CLAY2.nc  SAND2.nc  SILT2.nc
+
+documentation: Shangguan, W., Dai, Y., Duan, Q., Liu, B. and Yuan, H., 2014. A Global Soil Data Set for Earth System Modeling. Journal of Advances in Modeling Earth Systems, 6: 249-263 .
+
+data notes: 
+- horizontal resolution is ~global 30", 84N -> 56S
+- vertical resolution: 8 layers to the depth of 2.3 m:
+    0     - 0.045 m
+    0.045 - 0.091 m
+    0.091 - 0.166 m
+    0.166 - 0.289 m
+    0.289 - 0.493 m
+    0.493 - 0.829 m
+    0.829 - 1.383 m
+    1.383 - 2.296 m
+
+store 6 files in a directory named original/
+
+=======================================
+
+2. convert to texture using NCL; based on USDA soil texture triangle
+
+NCL does not handle the large size of this data so the process is split into 30deg x 30deg tiles
+
+The code produces a texture class for layers 0-1m and 1-2m.
+
+create a directory name tiles/
+
+since there is no data at either pole, run three separate scripts:
+
+convert_to_texture.ncl       : create tiles from 30S to 60N
+convert_to_texture_Spole.ncl : create tile  from 30S to 60S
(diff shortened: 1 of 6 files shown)


util/bnu_soil_gen/README
```diff
@@ -0,0 +1,144 @@
...
+
+4. final formatting to use in ufs_utils
+
+a. create 30s soil file template from viirs vegetation file
+
+cp VEG_LOCATION/vegetation_type.viirs.igbp.30s.nc soil_type.bnu.30s.nc
+
+b. run a bunch of nco commands to get the metadata and naming correct for soil type
+
+ncrename -O -h -v vegetation_type,soil_type soil_type.bnu.30s.nc
```
GeorgeGayno-NOAA, line 100: I don't see any of these attributes in the sample file on Hera. Did you not run this step yet?

barlage: All these commands have been run. Note that I'm taking the vegetation_type file as a template, most of these commands are removing metadata and attributes that came from the vegetation file since they are not relevant for the soil data.

barlage: Or in the case of the first command that you clipped, renaming the variable name to soil_type from vegetation_type

GeorgeGayno-NOAA: I understand now.  Thanks.

Could we add attributes to describe each soil category? I did not do that with the existing soil type data, but in retrospect I should have.

GeorgeGayno-NOAA:
Created a C768 grid using the new BNU soil type data file. Compared the soil type for each tile to their C768 OPS counterparts. Although, they used different input datasets, the general pattern is similar. Here are plots from tile 3:

BNU soil
BNU soil

OPS
OPS soil

util/bnu_soil_gen/README
barlage: Check this file:
hera:/scratch2/NCEPDEV/land/data/input_data/soil_type/soil_type.bnu.30s.nc

Let me know if those attributes are sufficient and I will update the readme with the nco commands. I also compressed the file.

barlage: I will also note that I am probably not doing this in the preferred way, which I assume would be to use some ufs_utils code to get the data file in the correct format. Since the 30" veg type data (vegetation_type.viirs.igbp.30s.nc) uses the same file format (esp. center and corner lat/lon), I'm cheating a bit to use a slew of nco commands to manipulate the file for soil type usage.

GeorgeGayno-NOAA: > Check this file: hera:/scratch2/NCEPDEV/land/data/input_data/soil_type/soil_type.bnu.30s.nc
> 
> Let me know if those attributes are sufficient and I will update the readme with the nco commands. I also compressed the file.

Perfect.

GeorgeGayno-NOAA: > I will also note that I am probably not doing this in the preferred way, which I assume would be to use some ufs_utils code to get the data file in the correct format. Since the 30" veg type data (vegetation_type.viirs.igbp.30s.nc) uses the same file format (esp. center and corner lat/lon), I'm cheating a bit to use a slew of nco commands to manipulate the file for soil type usage.

The codes I used to create all the files in ./sfc_climo are on Hera: /scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/climo_fields_netcdf/sorc. (directories ending in ./netcdf)

They were written before the UFS_UTILS repository existed and were never baselined. I should baseline one as a template for future users.

GeorgeGayno-NOAA (approved):
Looks good.

PR #766, opened 2023-01-30
Improve memory footprint in chgres_cube

### LINKED ISSUE #633: Memory bottleneck in chgres_cube
Users occasionally get out-of-memory issues when running chgres_cube for large domains. Almost always, this happens during the regridding of the 3-D winds to the edges of the grid box.

https://github.com/ufs-community/UFS_UTILS/blob/570ea3966c125ead7e90b4342be2efc22a8b1f41/sorc/chgres_cube.fd/atmosphere.F90#L347

I suspect this is because the ESMF field for winds is 4-dimensional (x,y,z wind components in the vertical). 

https://github.com/ufs-community/UFS_UTILS/blob/570ea3966c125ead7e90b4342be2efc22a8b1f41/sorc/chgres_cube.fd/atmosphere.F90#L632

Interpolating each wind component separately or as a field bundle would likely save memory.

GeorgeGayno-NOAA:
## DESCRIPTION OF CHANGES: 
Reduce memory by creating separate ESMF fields for the x,y, and z components of the wind, instead of one 4-D ESMF field that holds all three components. Memory becomes a problem when initializing C3072/L128, and more folks are running tests at this resolution.

There are two routines that convert winds with the same name. Give them unique names to prevent confusion.

Also move to ESMF v8.4, which improves processing of large data arrays.

## TESTS CONDUCTED: 
All chgres_cube consistency tests passed (as expected) on Jet, Orion, Hera, Cactus. The branch at 220ca48.

## DEPENDENCIES:
None.

## DOCUMENTATION:
N/A

## ISSUE: 
Fixes #633.


sorc/chgres_cube.fd/atmosphere.F90
```diff
@@ -343,8 +372,8 @@ subroutine atmosphere_driver(localpet)
  method=ESMF_REGRIDMETHOD_BILINEAR
 
  print*,"- CALL FieldRegridStore FOR 3D-WIND WEST EDGE."
- call ESMF_FieldRegridStore(wind_target_grid, &
-                            wind_w_target_grid, &
+ call ESMF_FieldRegridStore(xwind_target_grid, &
```
JeffBeck-NOAA, line 375: Looks like the print statement should be updated here.

sorc/chgres_cube.fd/atmosphere.F90
```diff
@@ -370,8 +415,8 @@ subroutine atmosphere_driver(localpet)
  method=ESMF_REGRIDMETHOD_BILINEAR
 
  print*,"- CALL FieldRegridStore FOR 3D-WIND SOUTH EDGE."
- call ESMF_FieldRegridStore(wind_target_grid, &
-                            wind_s_target_grid, &
+ call ESMF_FieldRegridStore(xwind_target_grid, &
```
JeffBeck-NOAA, line 418: Print statement needs to be updated.

sorc/chgres_cube.fd/atmosphere.F90
```diff
@@ -1465,8 +1625,19 @@ SUBROUTINE VINTG_WAM (YEAR,MONTH,DAY,HOUR)
 
 ! wind
  print*,"VINTG_WAM:- CALL FieldGet FOR 3-D ADJUSTED WIND."
- call ESMF_FieldGet(wind_target_grid, &
-                    farrayPtr=WIND2PTR, rc=rc)
+
```
JeffBeck-NOAA, line 1628: Should the print statement be updated here too?

util/gdas_init/config
```diff
@@ -52,16 +52,16 @@ EXTRACT_DATA=no
```
JeffBeck-NOAA: Why was this file changed?

sorc/chgres_cube.fd/atmosphere.F90
GeorgeGayno-NOAA: Done

sorc/chgres_cube.fd/atmosphere.F90
GeorgeGayno-NOAA: Done.

sorc/chgres_cube.fd/atmosphere.F90
GeorgeGayno-NOAA: Done.

util/gdas_init/config
GeorgeGayno-NOAA: Oh, I was probably testing something for another PR and it was inadvertently committed. I just reverted it.

JeffBeck-NOAA:
Thanks, @GeorgeGayno-NOAA!

JeffBeck-NOAA (approved):

Issue #768, opened 2023-02-05
additional resolutions should be available for cpld_gridgen


DeniseWorthen:
Currently the scripts to run the ``cpld_gridgen`` utility are limited to three ATM-OCN resolution pairs: C384 with 1/4deg ocean, C192 with 1/2deg ocean and C96 with 1deg ocean. Other resolution pairs are being used for testing by individuals or groups--for example C96 + 5 deg ocean, C48 + 5deg ocean, C192 + 1/4deg ocean. 

Baselines do not need to be maintained for these additional resolutions, but the user should be able to create their own desired combination using the cpld_gridgen utility.

@GeorgeGayno-NOAA Please assign me again. Thanks.

PR #769, opened 2023-02-05
Add additional resolutions for cpld_gridgen utilty

DeniseWorthen:
## DESCRIPTION OF CHANGES: 

- Updates the cpld_gridgen utility so that a user can generate a mix of ATM and OCN resolutions. The mapped ocean mask produced by this utility for a given ATM-OCN pair can then be used in the IC generation for the coupled model. 
- Removes the use of ``mastertask`` on print statements since the code runs in serial mode.
- Applies emacs standard indenting. This results in white space changes.

## TESTS CONDUCTED: 

Baselines have been run on wcoss2, hera, orion and jet at commit d4d0c3ebc1 and all cpld_gridgen baselines were B4B.

## DEPENDENCIES:

N/A

## DOCUMENTATION:

N/A

## ISSUE: 

- fixes issue #768


DeniseWorthen:
@GeorgeGayno-NOAA I've tested this on hera, orion and jet so far. On jet and orion the baselines reproduce. But on hera, the baselines fail because I added the 'netcdf4flag' setting to the creation of one of the files. However, if I make the same comparison (``nccmp -dmfqS``) outside of the rt.sh script, on hera it says they are identical. I'm not sure what is happening.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA I've tested this on hera, orion and jet so far. On jet and orion the baselines reproduce. But on hera, the baselines fail because I added the 'netcdf4flag' setting to the creation of one of the files. However, if I make the same comparison (`nccmp -dmfqS`) outside of the rt.sh script, on hera it says they are identical. I'm not sure what is happening.

If you remove the 'q' flag, you may get more information (I.e., nccmp -dmfS).

DeniseWorthen:
@GeorgeGayno-NOAA I determined that the different pass/fail results on hera vs the other platforms was user error. I also found a setting to add to the nccmp command (``-w format``) which allows rt.sh to ignore the netcdf4/classic format difference. 

Do I need to attach the RT logs to show each platform passed?

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA I determined that the different pass/fail results on hera vs the other platforms was user error. I also found a setting to add to the nccmp command (`-w format`) which allows rt.sh to ignore the netcdf4/classic format difference.
> 
> Do I need to attach the RT logs to show each platform passed?

That is not necessary. Just list which machines you tested on and the branch commit number you used.

DeniseWorthen:
@GeorgeGayno-NOAA I don't know why the CI test failed.

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA I don't know why the CI test failed.

I don't think the failure is anything you did. Sometimes tests just timeout. I will see if I can re-run it.

GeorgeGayno-NOAA:
@DeniseWorthen is this ready to be reviewed?

DeniseWorthen:
Yes, it is. Thanks. I put the information about the additional configurations in rt.conf. I don't know if that is sufficient.

ush/cpld_gridgen.sh
```diff
@@ -3,96 +3,104 @@ set -eux
...
+  export NPX=192
+elif [[ $MOSAICRES == C096 ]]; then
+  export MOSAICRES=C96
+  export NPX=96
+elif [[ $MOSAICRES == C048 ]]; then
+  export MOSAICRES=C48
+  export NPX=48
+fi
+if [ $RESNAME = 500 ]; then
+  export FIXDIR_PATH=/scratch1/NCEPDEV/global/glopara/fix/mom6/20220805/500/
```
GeorgeGayno-NOAA, line 45: This looks like a Hera directory. How does this work on another machine?

DeniseWorthen: I wasn't sure if you wanted to add the required 5deg ocean files to the usual location since we're not running baselines on for the 5deg case. There is no use-case really for the 5-deg ocean other than the GW infrastructure work Rahul is doing.

GeorgeGayno-NOAA: Ok. Does this code have a large user base or is it just you right now? If the former, do you want to add a comment about running the 5-deg on Hera only?

GeorgeGayno-NOAA:
Checked out branch at d4d0c3e on Hera, Orion, Jet and Dogwood. Compiled it and successfully ran the regression tests. On Hera, the doxygen was generated with no warnings and the unit tests ran successfully.

ush/cpld_gridgen.sh
DeniseWorthen: One of the reasons I'm making this PR is so that other users can generate their own configurations. Right now, they simply ask me to generate what they need for them. Once we get this committed, I'll try steering users to ufs-utils instead. 

I will add a note that 5-deg is only on hera.

GeorgeGayno-NOAA: Ok.

GeorgeGayno-NOAA:
For a recent HAFS implementation, we got in trouble for having 'goto' statements in one of our codes. I did not see any in your code. But can you double check?

DeniseWorthen:
I have no 'goto' statements in the cpld_gridgen source code.

GeorgeGayno-NOAA:
@DeniseWorthen Unless you have any final changes, I can merge this afternoon.

DeniseWorthen:
It should be good to go. Thanks.

GeorgeGayno-NOAA (approved):

Issue #596, opened 2021-11-05
FVCOM_tools fails when using GNU 


dmwright526:
FVCOM_tools and the corresponding unit test creates a segfault when running the code using the GNU compilers. Code was originally tested and developed using Intel compilers, so the exact reason for the error is unclear. The segfault points to the fcst_nwp object and potentially one of the pointers within the object.

GeorgeGayno-NOAA:
@GeorgeVandenberghe-NOAA I know you are very busy, but could we get a bit of your time? @dmwright526 is getting an odd seg fault when running with GNU and we are looking for some guidance. I will let him explain the problem.

kgerheiser:
I would recommend compiling and running with debug flags `-g -fbacktrace -fcheck=bounds` to see if it catches anything. And probably worth it to try with Intel too `-g -traceback -check all`.

kgerheiser:
Trying to build the tests, but the function call is missing some arguments:

```
  125 |  call fcst%read_n(trim(fvcomfile),' FVCOM',wcstart,lbclon,lbclat,lbctimes,t2,lbcmask,lbcsst,lbcice,lbcsfcT,lbciceT,fv3sfcTl)
      |                                                                                                                            1
Error: Missing actual argument for argument 'zorl' at (1)
```

And compared to the function defintion:

```
 read_nwp(this,filename,itype,wcstart,numlon,numlat,numtimes,time_to_get,mask,sst,ice,sfcT,iceT,sfcTl,zorl,hice)
```

Looks like it's missing 2 arguments: `zorl` and `hice`.

dmwright526:
@kgerheiser Use this hash instead:
https://github.com/ufs-community/UFS_UTILS/commit/2a08859a1a591e3e8297042710bd1a9bf522b6a3

This was before the recent update we had to push through quickly due to wrong initialization of surface roughness length causing the MYNN scheme to crash. It will take me a bit to update the tests and dummy data to include the fields, so this will at least allow to see if you can diagnose the problem.

kgerheiser:
Thanks, I was able to build and reproduce the problem. Intel debug flags didn't catch anything.

GeorgeGayno-NOAA:
> Thanks, I was able to build and reproduce the problem. Intel debug flags didn't catch anything.

@kgerheiser We know you are super busy. Any progress on this? Should we ask @arunchawla-NOAA for some of your time?

edwardhartnett:
[...]

PR #790, opened 2023-03-17
Remove dependency on module_nwp_base.f90

DavidWright-NOAA:
## DESCRIPTION OF CHANGES: 
Current fvcom_tools does not work with GNU compiler (as stated in issue #596 ). Code previously resulted in a seg fault with the fcst_nwp object in the process_FVCOM.f90 file when using GNU (but not Intel). Fix presented here removes the use of pointers in the code base (specifically pointers to NULL()) and removes the need for module_nwp_base.f90. The file module_nwp_base.f90 has been removed completely along with the use of it in the CMake file to compile. The unit test for fvcom_tools has also been uncommented in tests/CMakeFile.

## TESTS CONDUCTED: 
Code was tested with the unit test for both Intel and GNU compilers and both passed. Further testing was done with a cold start surface file (sfc_data.nc) from UFS-SRW and personal fvcom files and behavior was as expected for GNU compiler.

## DEPENDENCIES:
No dependencies.

## DOCUMENTATION:
No documentation should need to be updated (to my knowledge) since the functionality remains the same.

## ISSUE: 
Resolves issue #596 

## CONTRIBUTORS (optional): 
@GeorgeGayno-NOAA provided some initial debugging of the code.


JeffBeck-NOAA:
@DavidWright-NOAA, thanks for tracking this down!  Did your cold-start tests result in identical sfc_data.nc files between Intel and GNU?

DavidWright-NOAA:
@JeffBeck-NOAA I just did an ncdiff on both .nc files (GNU and Intel generated) and they were identical in their outputs.

JeffBeck-NOAA:
> @JeffBeck-NOAA I just did an ncdiff on both .nc files (GNU and Intel generated) and they were identical in their outputs.

Thanks, @DavidWright-NOAA!

JeffBeck-NOAA (approved):

GeorgeGayno-NOAA:
Thanks @DavidWright-NOAA. I will merge your changes.

PR #808, opened 2023-04-10
Fix bug in chgres_cube subroutine search_many

### LINKED ISSUE #797: chgres_cube - problem with call to routine "search_many"
When running is Debug mode, the chgres_cube regression tests are crashing at the calls to routine "search_many". That routine is called from each MPI task, but some of the arguments have only been allocated on task 0. For example, the tests crash with an error about 'land_target_one_tile' not being allocated. 

```
   if (localpet == 0) then
     allocate(land_target_one_tile(i_target,j_target))
     land_target_one_tile = 0
     where(mask_target_one_tile == 1) land_target_one_tile = 1
   endif

   call search_many(num_fields,bundle_allland_target,data_one_tile, land_target_one_tile,&
                    tile,search_nums,localpet)

   if (localpet == 0) deallocate(land_target_one_tile)
```

After a code handoff, NCO will now compile/run all tags in Debug mode. For the Intel compiler, that mode must include at least `-check all -ftrapuv`.

As a future item, at least one of the regression tests should run in Debug mode.

LarissaReames:
## DESCRIPTION OF CHANGES: 
There was a bug with an unallocated array being used on all PETs in a call to search_many when it only needed to be used on the head node. I fixed this is in all calls to search_many subroutine, including the unit test for this routine. 

## TESTS CONDUCTED: 
Consistency and unit tests, with code compiled in debug mode, passed on Jet.

## DEPENDENCIES:
N/A

## DOCUMENTATION:
No documentation changes needed.

## ISSUE: 
Fixes #797 

## CONTRIBUTORS (optional): 
N/A


LarissaReames:
All of the tests appear to be failing in places that have nothing to do with anything I've changed, so I'm not really sure how to rectify that. This certainly goes back to our discussions in our last management meeting about having so many tests that constantly need updating but that aren't adding much value.

It looks like all of the GitHub actions still use nceplibs which we no longer use for modules, instead using hpc-stack, so some library versions required by UFS_UTILS aren't available. Should we create an issue to update the actions to source and build hpc-stack instead?

GeorgeGayno-NOAA:
> All of the tests appear to be failing in places that have nothing to do with anything I've changed, so I'm not really sure how to rectify that. This certainly goes back to our discussions in our last management meeting about having so many tests that constantly need updating but that aren't adding much value.
> 
> It looks like all of the GitHub actions still use nceplibs which we no longer use for modules, instead using hpc-stack, so some library versions required by UFS_UTILS aren't available. Should we create an issue to update the actions to source and build hpc-stack instead?

Don't worry about the failing tests. They are failing for reasons other than errors in our codes.  @arunchawla-NOAA is aware of the problem.

GeorgeGayno-NOAA:
Does argument `field_data_2d` need to be an argument of `search_many`? The routine does a FieldGather and FieldScatter on that array. Can it be declared locally? Because the routine expects `field_data_2d` to be dimensioned as `i_target` and `j_target`. And on most tasks, that array is dimensioned as (0,0). I wonder if that would cause problems.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -851,11 +851,12 @@ subroutine interp(localpet)
    if (localpet == 0) then   
      where(mask_target_one_tile == 1) mask_target_one_tile = 0
      where(mask_target_one_tile == 2) mask_target_one_tile = 1
+     call search_many(num_fields,bundle_seaice_target,tile, search_nums,localpet, &
```
GeorgeGayno-NOAA, line 854: Compilation is failing here because field_data_3d is no longer an argument.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -1202,8 +1207,13 @@ subroutine interp(localpet)
     if(ESMF_logFoundError(rcToCheck=rc,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__)) &
        call error_handler("IN FieldGather", rc)
 
-   call search_many(num_fields,bundle_landice_target,data_one_tile, land_target_one_tile,& 
-                    tile,search_nums,localpet,terrain_land=data_one_tile2,field_data_3d=data_one_tile_3d)
+   if (localpet==0) then
+      call search_many(num_fields,bundle_landice_target,tile,search_nums,localpet,&
+                terrain_land=data_one_tile2,mask=land_target_one_tile)
+   else
+      call search_many(num_fields,bundle_landice_target,tile,search_nums,localpet,&
```
GeorgeGayno-NOAA: I think you can remove terrain_land from the argument list for all tasks > 0.

sorc/chgres_cube.fd/surface.F90
LarissaReames: Yes I'm working on debugging. I made changes on one system and wanted to get it to another to do some more work, so that's what that commit was. I've got more coming soon.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -1416,9 +1426,13 @@ subroutine interp(localpet)
    call ESMF_FieldGather(soil_type_target_grid, data_one_tile2, rootPet=0,tile=tile, rc=rc)
    if(ESMF_logFoundError(rcToCheck=rc,msg=ESMF_LOGERR_PASSTHRU,line=__LINE__,file=__FILE__))&
       call error_handler("IN FieldGather", rc)
-      
-   call search_many(num_fields,bundle_nolandice_target,data_one_tile, mask_target_one_tile,& 
-                    tile,search_nums,localpet,soilt_climo=data_one_tile2, field_data_3d=data_one_tile_3d)
+   if (localpet==0) then 
+      call search_many(num_fields,bundle_nolandice_target,tile,search_nums,localpet, &
+                soilt_climo=data_one_tile2, mask=mask_target_one_tile)
+   else
```
GeorgeGayno-NOAA, line 1431: Here as well. Remove soilt_climo from the call for tasks > 0.

LarissaReames: Removed in my latest commit.

sorc/chgres_cube.fd/surface.F90
LarissaReames: Removed in my latest commit.

sorc/chgres_cube.fd/surface.F90
```diff
@@ -3301,9 +3313,8 @@ end subroutine regrid_many
 !! @param[in]  soilt_climo  (optional) A real array size i_target,j_target of climatological soil type on the target grid
```
GeorgeGayno-NOAA, line 3310: Don't forget to adjust the doxygen block.

GeorgeGayno-NOAA:
Checked out the branch at 36fc338 on WCOSS2. The doxygen built with no warnings. All regression tests passed. 

Then, recompiled the branch using 'Debug' mode. All regression tests completed without error. There were very small differences with the baseline data, which were likely due to the different compiler options used in 'Debug' mode. Importantly, the fix to `search_many` is working.

GeorgeGayno-NOAA (approved):

PR #873, opened 2023-11-20
Update chgres_cube for fractional grids

### LINKED ISSUE #123: Update chgres_cube for mixed land/non-land points
The grid generation is being updated to output lake fraction and lake depth (#111).  As a result, some model grid points will be a mix of land and non-land.  Currently, `chgres_cube` assumes that points can only be land or non-land, not both.   Update `chgres_cube` accordingly.

GeorgeGayno-NOAA:
## DESCRIPTION OF CHANGES: 
Update `chgres_cube` for fractional (partial land/non-land) grids. Update the format of the surface coldstart file (called v2). These changes are designed to be backwards compatible with the forecast model.

## TESTS CONDUCTED: 

- [x] Compile branch on all Tier 1 machines using Intel (Orion, Jet, Hera and WCOSS2). Done using 45ca85a.
- [x] Compile branch on Hera using GNU. Done using 45ca85a.
- [x] Compile branch in 'Debug' mode on WCOSS2. Done using 45ca85a
- [x] Run unit tests locally on any Tier 1 machine. Done on dogwood using 45ca85a. All passed.
- [x] Run relevant consistency tests locally on all Tier 1 machine. Done using 45ca85a. All `chgres_cube` tests failed as expected. Most `grid_gen` tests failed as expected. For details, see notes below.

Describe any additional tests performed.

## DEPENDENCIES:
https://github.com/NOAA-EMC/fv3atm/pull/698
https://github.com/ufs-community/ufs-weather-model/pull/2005

## DOCUMENTATION:
N/A

## ISSUE: 
Fixes #123.


BrianCurtis-NOAA:
@GeorgeGayno-NOAA If your FV3 PR NOAA-EMC/fv3atm#698 is ready to go, should this still be in draft form?

reg_tests/chgres_cube/25km.conus.gfs.grib2.sh
```diff
@@ -12,10 +12,10 @@ set -x
 export DATA=$OUTDIR/25km_conus_gfs_grib2
 rm -fr $DATA
 
-export CRES=403
+export CRES=405
 export KMRES=25km
-export FIXfv3=${HOMEreg}/fix/RRFS_CONUS_${KMRES}
-export FIXsfc=${FIXfv3}/fix_sfc
+export FIXfv3=${HOMEreg}/fix.new/RRFS_CONUS_${KMRES}
+
```
JeffBeck-NOAA, line 18: Is there a reason the fix directory can't stay as "fix", and the old directory be renamed?

reg_tests/chgres_cube/25km.conus.gfs.grib2.sh
```diff
@@ -24,7 +24,7 @@ export VARMAP_FILE=${HOMEufs}/parm/varmap_tables/GSDphys_noaero_var_map.txt
 export INPUT_TYPE='grib2'
 export CONVERT_NST=".false."
 export THOMPSON_AEROSOL_FILE=${HOMEufs}/fix/am/Thompson_MP_MONTHLY_CLIMO.nc
-export OROG_FILES_TARGET_GRID="C403_oro_data.tile7.halo4.nc"
+export OROG_FILES_TARGET_GRID="C405_oro_data.tile7.nc"
 export REGIONAL=1
```
JeffBeck-NOAA, line 28: What was the reason for the minor C-resolution change for these files?

reg_tests/chgres_cube/c192.fv3.history.sh
```diff
@@ -13,17 +13,18 @@ export DATA=$OUTDIR/c192_fv3_history
 rm -fr $DATA
 
 export CRES=192
-export FIXfv3=${HOMEreg}/fix/C192
-export FIXsfc=${FIXfv3}/fix_sfc
+export ocn=100
+export FIXfv3=${HOMEreg}/fix/C${CRES}
+
```
JeffBeck-NOAA, line 18: Does this also need to point to "fix.new"?  It'd be best if the new files could be populated to "fix", and the old directory renamed instead.

reg_tests/chgres_cube/c192.gfs.grib2.sh
```diff
@@ -13,8 +13,9 @@ export DATA=$OUTDIR/c192_gfs_grib2
 rm -fr $DATA
 
 export CRES=192
-export FIXfv3=${HOMEreg}/fix/C192
-export FIXsfc=${FIXfv3}/fix_sfc
+export ocn=100
+export FIXfv3=${HOMEreg}/fix/C${CRES}
+
```
JeffBeck-NOAA, line 18: All these files still point to "fix" and not "fix.new".

reg_tests/chgres_cube/c96.regional.sh
JeffBeck-NOAA, line 1: Why delete the regional test?

sorc/chgres_tools.fd/check.F90
JeffBeck-NOAA, line 1: Can this file name be a bit more descriptive?

sorc/chgres_tools2.fd/CMakeLists.txt
JeffBeck-NOAA, line 1: Did these "chgres.tools2.fd" files get committed accidentally?

JeffBeck-NOAA (changes requested):

reg_tests/chgres_cube/25km.conus.gfs.grib2.sh
GeorgeGayno-NOAA: I switched it back. I was just testing something.

sorc/chgres_tools.fd/check.F90
GeorgeGayno-NOAA: A temporary utility program that I will delete.

sorc/chgres_tools2.fd/CMakeLists.txt
GeorgeGayno-NOAA: Another temporary program that I will delete

reg_tests/chgres_cube/25km.conus.gfs.grib2.sh
GeorgeGayno-NOAA: This version of chgres requires that all surface climo fields (like veg type) be defined at all points with at least some land (instead of 50% land). So, I had to regenerate all the fixed data for the regression tests. Using what I thought are the grid specs for the regional grids, I was unable to get the CRES value to match. After a few attempts, I ended up using the closest one. The original files were generated three years ago, so code and library changes are probably a factor. But the new files are very close to the original ones. If you visualize both, they look like the identical grids.

reg_tests/chgres_cube/c96.regional.sh
GeorgeGayno-NOAA: The C96 test was the first regional regression test. It uses the GFDL regional grid definition, not the new regional ESG method. Since then, you and Larissa added several new regional tests for the GRIB2 option. Since your new tests exercise the regional function (creation of LBC file, stripping of halo region), I thought this C96 test is now redundant. 

You work with the regional group. If this is an important configuration for you, I can add back this test.

reg_tests/chgres_cube/25km.conus.gfs.grib2.sh
JeffBeck-NOAA: Thanks!

reg_tests/chgres_cube/c96.regional.sh
JeffBeck-NOAA: Yep, we're fine getting rid of it, particularly since it was the old regional grid method.

JeffBeck-NOAA (approved):

GeorgeGayno-NOAA:
All `grid_gen` consistency tests failed except the c96 uniform test. 

The `sfc_climo_gen` code was updated to always map surface fields to any point with at least some land (this is expected by updated `chgres_cube`). Previously, surface fields were mapped to points with at least some land only when the 'add_lake' option was used. Otherwise, surface fields were only mapped to points with 50% land. 

The c96 uniform tests used the 'add_lake' option (and passed). The others did not use that option. As a result, the surface files from the remaining tests differed from the baseline (the 'grid' and 'oro' files were the same). As expected, the surface files created by the branch included additional data points at coasts. Inland, the data were the same.

GeorgeGayno-NOAA:
A note about the `chgres_cube` consistency tests. 

Since the updated `chgres_cube` requires valid surface values at all points with at least some land, the 'fixed' data input to the tests was updated.

The c96 and c192 tests now use the '202310127' version files ('grid', 'oro' and 'sfc' files).

The RRFS 13KM and 25KM fixed files ('grid', 'oro' and 'sfc' files) were recreated using the branch. Despite several attempts, I could not get the grids to exactly match (the CRES is slightly different). However, they are very close. 

The RRFS 3KM surface fixed files were updated, but the same 'grid' and 'oro' files were used.

GeorgeGayno-NOAA:
All `chgres_cube` consistency tests failed as expected. 

Failures were due to: 

- The new chgres outputs a different surface coldstart file format
- Differences in the input fixed data (new 'grid', 'oro' and 'sfc' files)
- The new chgres outputs data for a fractional land/non-land grid.

Files from the branch tests were visualized against the baseline data. The differences I saw were consistent with the updates for this PR and did not indicate a problem with the branch.

PR #998, opened 2024-11-18
CI fixes, Nov 24

AlexanderRichert-NOAA:
## DESCRIPTION OF CHANGES: 
This PR fixes some issues with the ubuntu_intel CI. After getting around the initial issue (which is that it was using GCC rather than oneAPI), I ran into a couple more problems. The weirdest was Python 3.13 failing to install through Spack. I have a suspicion that it has to do with the miniconda installation that the runner is loading into the environment, but in any case reverting to Python 3.11 fixed it. [Eventually](https://github.com/spack/spack/pull/45474), the logic for incorporating the external intel-oneapi-mpi should be much cleaner.

## TESTS CONDUCTED: 
Ran in [CI](https://github.com/AlexanderRichert-NOAA/UFS_UTILS/actions/runs/11901480143/job/33164517452).

## DEPENDENCIES:
none

## DOCUMENTATION:
n/a

## ISSUE: 
none

### DIFF
--- a/.github/workflows/ubuntu_intel.yaml
+++ b/.github/workflows/ubuntu_intel.yaml
@@ -67,6 +67,7 @@ jobs:
           sudo apt install cmake
           spack external find
           spack add intel-oneapi-mpi
+          spack config add "packages:all:prefer:'%oneapi'"
           cat $SPACK_ENV/spack.yaml
           spack concretize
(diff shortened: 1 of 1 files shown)


GeorgeGayno-NOAA (approved):
Thanks!

PR #1003, opened 2024-12-04
ocean_merge - Increase directory path variable length

### LINKED ISSUE #1002: Increase path variable length in ocean_merge
A recent `grid_gen` regression test failure on Hercules (under the cronned role account) was due to insufficiently long character variable declarations.  The test that failed was the c96.uniform test, which is the only one that runs the `ocean_merge` step.

GeorgeGayno-NOAA:
## DESCRIPTION OF CHANGES: 
Some of the path variables do not have enough characters. 

## TESTS CONDUCTED: 

- [x] Compile branch on all Tier 1 machines using Intel (Orion, Jet, Hera, Hercules and WCOSS2). Done using c1a84df.
- [x] Compile branch on Hera using GNU. Done using c1a84df.
- [x] Compile branch in 'Debug' mode on WCOSS2. Done on Dogwood using c1a84df.
- [x] Run unit tests locally on any Tier 1 machine. Done on Hera (with Gnu) using c1a84df.
- [x] Run `grid_gen` consistency tests locally on all Tier 1 machines. Done using c1a84df. All tests passed as expected.

## DEPENDENCIES:
None.

## DOCUMENTATION:
N/A

## ISSUE: 
Fixes #1002.

### DIFF
--- a/sorc/ocean_merge.fd/merge_lake_ocnmsk.F90
+++ b/sorc/ocean_merge.fd/merge_lake_ocnmsk.F90
@@ -18,8 +18,8 @@ program merge_lake_ocnmsk
 
   implicit none
 
-  character(len=120) :: pth1
-  character(len=120) :: pth2,pth3
+  character(len=200) :: pth1
+  character(len=200) :: pth2,pth3
   character(len=10)  :: atmres,ocnres
 
   integer :: binary_lake
--- a/sorc/ocean_merge.fd/namelist.F90
+++ b/sorc/ocean_merge.fd/namelist.F90
@@ -16,9 +16,9 @@ subroutine read_nml(ocean_mask_dir, lake_mask_dir, atmres,ocnres,out_dir,binary_
 
   integer :: unit=7, io_status
 
-  character(len=120), intent(out) :: ocean_mask_dir
-  character(len=120), intent(out) :: lake_mask_dir
-  character(len=120), intent(out) :: out_dir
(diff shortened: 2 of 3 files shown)


sorc/ocean_merge.fd/merge_lake_ocnmsk.F90
```diff
@@ -18,8 +18,8 @@ program merge_lake_ocnmsk
 
   implicit none
 
-  character(len=120) :: pth1
-  character(len=120) :: pth2,pth3
+  character(len=200) :: pth1
+  character(len=200) :: pth2,pth3
```
sanatcumar, line 22: I see different values of  atleast 1,2,3,4,6,20 ,32,100,128160,256,512 in different codes for the len=.
Is there a need to standardize or extend them as we need?

sanatcumar (approved):
Compiled c1a84df, no issues noticed.

sorc/ocean_merge.fd/merge_lake_ocnmsk.F90
GeorgeGayno-NOAA: I don't know if I would standardize length of character strings. Rather, when we review future PRs, we should keep in mind that any string used for directory paths should be 'large'.

GeorgeGayno-NOAA:
@sanatcumar thanks for your prompt review.

PR #1057, opened 2025-05-08
Add DATM mesh generation capability 

### LINKED ISSUE #1056: Add creation of DATM meshes
The UWM has several DATM configurations. Each DATM domain needs a mesh file which matches the domain of the forcing data. There are two issues w/ the DATM meshes currently in UWM:

1) There is no record of how they were produced
2) They contain a land-mask, which prevents use of bilinear mapping of DATM states to other components (ufs-community/ufs-weather-model/issues/2686)

The required meshes are generated from a SCRIP file using ``ESMF_Scrip2Unstruct``. The SCRIP files themselves are generated using NCO's ``ncremap`` command.  

The generation of the DATM meshes can be added as a new script in the util subdirectory.

DeniseWorthen:
## DESCRIPTION OF CHANGES: 

Currently the meshes used by the DATM component of UWM have no known, documented creation method. These meshes are fixed files and should have a known, reproducible method to create them. This PR adds the creation of the DATM meshes used by UWM to the cpld_gridgen utility. This capability is added here even though it is not dependent on any ocean resolution or even the compiling of the cpld_gridgen executable. However, it does allow the creation of the DATM meshes in a documented manner.

The capability is added through a new rt.sh option (-d) and rt.conf member ``datm | -99``. In combination, these options allow bypassing the build of the cpld_gridgen executable and instead rt.sh simply runs a series of commands to first generate the SCRIP file for the domain and then convert it to an ESMF mesh.

## TESTS CONDUCTED: 

- [ ] Compile branch on all Tier 1 machines using Intel (Orion, Jet, Hera, Hercules and WCOSS2).
- [ ] Compile branch on Hera using GNU.
- [ ] Compile branch in 'Debug' mode on WCOSS2.
- [ ] Compile with Doxygen on any machine with no errors.
- [ ] Run unit tests locally on any Tier 1 machine.
- [ ] Run relevant consistency tests locally on all Tier 1 machines.

Describe any additional tests performed.

The DATM meshes produced have been compared to those currently in use by UWM. The differences in centerCoord are all on the order of ~10-17, the differences in the nodeCoords are all on the order of ~10-14 and the differences in the element areas are all ~10-12. The DATM meshes produced also have a uniform elementMask==1, as required by the switch to bilinear mapping.

A baseline run was made on Hercules with the ``datm`` commented out in the rt.conf. All tests pass.

The relevant UWM RTs which will use the new meshes have been run and compared to what is currently in use. The meshes are behaving as expected. See the UWM issue for figures.

## ISSUE: 
- fixes #1056 
- required by ufs-community/ufs-weather-model/issues/2686


DeniseWorthen:
@GeorgeGayno-NOAA I've run on both Hera and Hercules, and the existing cpld_gridgen tests all run and compare B4B. Since this PR doesn't add any compiled code, I wasn't sure which testing I needed to perform.

DeniseWorthen:
@GeorgeGayno-NOAA I'm having second thoughts about producing the required DATM meshes in this way. I wonder whether it would be acceptable to just add a script to the glopara/fix/datm directories which can produce the meshes if required. I think adding it to cpld_gridgen works, but it might be overkill?

GeorgeGayno-NOAA:
> @GeorgeGayno-NOAA I'm having second thoughts about producing the required DATM meshes in this way. I wonder whether it would be acceptable to just add a script to the glopara/fix/datm directories which can produce the meshes if required. I think adding it to cpld_gridgen works, but it might be overkill?

They may not allow source code in the datm directories. You can consider adding the script to the UFS_UTILS utility directory - https://github.com/ufs-community/UFS_UTILS/tree/develop/util

DeniseWorthen:
Ah, yes I think that makes sense. I'll try to add the scripts there. Thanks!

GeorgeGayno-NOAA:
> Ah, yes I think that makes sense. I'll try to add the scripts there. Thanks!

Will this utility be widely used? If so, create a driver script for all supported machines. If not, create scripts for only the necessary machines.

DeniseWorthen:
The files produced by the script will likely never need to be regenerated, although new meshes may be required. The thing I want to be sure of is that we know how these meshes are generated.

There are existing meshes in the glopara/fix/datm dated from 2022. I don't know who or how those meshes were generated, but the fact they contain a land mask is a problem. 

Eventually, G-W needs to be able to build and run the NG-GODAS application and that is when these files will be needed. For now, I only really need to generate them on Hera and add them to the UWM Regression test input-data directory on Hera.

DeniseWorthen:
Closing in preference to #1060

PR #1063, opened 2025-05-27
Add time dimension to sfc regrid based on namelist specification

tsga:
- This PR modifies the surface regridding code to add time dimensions in output files based on namelist specification. The previous version added time dimension only to files with more than one timesteps, but the UFS model expects time dim for all files. 
- This PR also adds a surface regridding namelist template (used in Global Workflow)
- For inputs with multiple timesteps, separate land mask specific to each timestamp is applied.

## TESTS CONDUCTED: 

- [x] Compile branch on all Tier 1 machines using Intel (Orion, Jet, Hera, Hercules and WCOSS2). Done using a2629eb.
- [x] Compile branch on Hera using GNU. Done using a2629eb.
- [x] Compile branch in 'Debug' mode on WCOSS2. Successfully done using a2629eb. 
- [x] Compile with Doxygen on any machine with no errors. Done on WCOSS2 using a2629eb.
- [x] Run unit tests locally on any Tier 1 machine. Done on Hera using GNU and a2629eb.
- [x] Run `regrid_sfc` consistency tests locally on all Tier 1 machines. Done using a2629eb. See note below.

Note on the consistency tests. This PR adds a time dimension to the output files. Therefore, new baselines were generated and placed in the official locations. The tests were run against this new baseline (and passed).

## DEPENDENCIES:
None.

## DOCUMENTATION:
The Doxygen was built and visualized. The `regrid_sfc` section looked good.

## ISSUE:
None.

### DIFF
--- a/parm/regrid_sfc/regrid.nml_tmpl
+++ b/parm/regrid_sfc/regrid.nml_tmpl
@@ -1,0 +1,28 @@
+&config
+  n_vars=@[n_vars],
+  variable_list=@[soil_incr_vars]
+  missing_value=0.,
+  time_list=@[time_list],
+  add_time_dim=@[add_time_dim], 
+/
+&input
+  gridtype="gau_inc",
+  ires=@[ires],
+  jres=@[jres],
+  fname=@[in_fname], 
+  dir="./",
+  fname_coord="gaussian_scrip.nc",
+  dir_coord="./",
+  fname_mask=@[fname_mask_in],
+  dir_mask=@[dir_mask_in]
+/
+&output
+  gridtype="fv3_rst",
+  ires=@[ireso],
+  jres=@[jreso],
+  fname=@[out_fname],
+  dir="./",
+  fname_mask="vegetation_type" 
+  dir_mask="./"
+  dir_coord="./"
+ /
--- a/reg_tests/regrid_sfc/gauss2fv3incr.sh
+++ b/reg_tests/regrid_sfc/gauss2fv3incr.sh
@@ -57,6 +57,7 @@ cat << EOF > regrid.nml
   variable_list="soilt1_inc", "soilt2_inc", "slc1_inc", "slc2_inc",
   missing_value=0.,
   time_list=6,
+  add_time_dim=.true.,
   extrap_levs=2,
  /
  &input
--- a/reg_tests/regrid_sfc/gauss2fv3incr.sh
+++ b/reg_tests/regrid_sfc/gauss2fv3incr.sh
@@ -34,7 +34,7 @@ mkdir -p $DATA
 cd ${DATA}
 
 # input, increment
-/bin/cp -p ${COMIN_REGTEST}/sfcincr_gsi ${DATA}/sfcincr_gsi
+/bin/cp -p ${COMIN_REGTEST}/sfcincr_gsi ${DATA}/sfcincr_gsi006.nc
 
 # input, fixed files
 ln -sf "${FIXorog}/gaussian.${LONB_CASE_IN}.${LATB_CASE_IN}.nc" \
@@ -56,6 +56,7 @@ cat << EOF > regrid.nml
   n_vars=4,
   variable_list="soilt1_inc", "soilt2_inc", "slc1_inc", "slc2_inc",
   missing_value=0.,
+  time_list=6,
   extrap_levs=2,
  /
  &input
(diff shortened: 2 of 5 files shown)


tsga:
tagging @ClaraDraper-NOAA @CoryMartin-NOAA

CoryMartin-NOAA (approved):
I did not test but changes look reasonable

GeorgeGayno-NOAA:
The regression test failed on WCOSS2 with this error:

```
 - FATAL ERROR:
 n_tims < 1. must have at least one valid increment hour in time_list
 - IOSTAT IS:            1
```

tsga:
@GeorgeGayno-NOAA @ClaraDraper-NOAA I am trying to run the regression tests on hera and wcoss. In both platforms, I can't see the gaussian increment files in /scratch1/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/regrid_sfc/ (hera) and /lfs/h2/emc/nems/noscrub/emc.nems/UFS_UTILS/reg_tests/regrid_sfc

tsga:
@GeorgeGayno-NOAA sorry about the oversight about testing. On hera, I ran "reg_tests/regrid_sfc/driver.sh" and got in summary.log
consistency.log01:<<< REGRID SFC TEST PASSED >>>
and 
image

tsga:
WCOSS regression tests: <<< REGRID SFC TEST PASSED >>>

GeorgeGayno-NOAA:
@tsga - in the future, open an issue first.

GeorgeGayno-NOAA:
@tsga - When I look at the output from the regression test, I do not see a time dimension. Is that intentional? Do you want to add another test for the time dimension option?

tsga:
> @tsga - When I look at the output from the regression test, I do not see a time dimension. Is that intentional? Do you want to add another test for the time dimension option?

Yes, the default option is the no-time-dimension. I wonder if it is worth having to run two tests as this is a minor difference. @ClaraDraper-NOAA what do you think?

GeorgeGayno-NOAA:
The Doxygen build failed with this error:

```
/scratch1/NCEPDEV/da/George.Gayno/ufs_utils.git/UFS_UTILS.tsga/sorc/regrid_sfc.fd/grids_IO.F90:48: error: The following
parameters of grids_io::setup_grid(integer, intent(in) localpet, integer, intent(in) npets, type(grid_setup_type), 
intent(in) grid_setup, type(esmf_grid), intent(out) mod_grid, integer, intent(in), optional timestamp) are not documented:
  parameter 'timestamp' (warning treated as error, aborting now)
```

The 'timestep' argument has not been defined.

GeorgeGayno-NOAA:
> > @tsga - When I look at the output from the regression test, I do not see a time dimension. Is that intentional? Do you want to add another test for the time dimension option?
> 
> Yes, the default option is the no-time-dimension. I wonder if it is worth having to run two tests as this is a minor difference. @ClaraDraper-NOAA what do you think?

If it is minor, then it is not mandatory to create another test. But I will defer to you.

tsga:
> > > @tsga - When I look at the output from the regression test, I do not see a time dimension. Is that intentional? Do you want to add another test for the time dimension option?
> > 
> > 
> > Yes, the default option is the no-time-dimension. I wonder if it is worth having to run two tests as this is a minor difference. @ClaraDraper-NOAA what do you think?
> 
> If it is minor, then it is not mandatory to create another test. But I will defer to you.

I just talked to Clara. She prefers the files with time dimension to be default in the test case. I have updated the test accordingly, but Clara may have more comments soon.

GeorgeGayno-NOAA:
> > > > @tsga - When I look at the output from the regression test, I do not see a time dimension. Is that intentional? Do you want to add another test for the time dimension option?
> > > 
> > > 
> > > Yes, the default option is the no-time-dimension. I wonder if it is worth having to run two tests as this is a minor difference. @ClaraDraper-NOAA what do you think?
> > 
> > 
> > If it is minor, then it is not mandatory to create another test. But I will defer to you.
> 
> I just talked to Clara. She prefers the files with time dimension to be default in the test case. I have updated the test accordingly, but Clara may have more comments soon.

Ok. I will host new baseline data on each machine.

GeorgeGayno-NOAA (approved):
Looks good. Will merge.

GeorgeGayno-NOAA:
@tsga - anything else before I merge?

tsga:
> @tsga - anything else before I merge?

No I think it is good. Thank you @GeorgeGayno-NOAA !

Issue #1078, opened 2025-07-11
Add check in chgres_cube to catch incorrect specification of the input orography file


ClaraDraper-NOAA:
when running chgres_cube with input restart files, it is possible to specify an incorrect input orography file (as in, not the same orography file that was used to create the input restarts). This will then produce a faulty output restart that can then crash due to mismatch between the model states and the land mask (veg type, etc). 

We need to add a check to the chgres_cube that ensures that the input restart and specified orography file have matching land / non-land grid cells. For example, something like: 

for veg_type in the restart and land_frac in the orog file: 
- check that  ( veg_type gt 0) or ( veg_type ne water_type )  if and only if ( land_frac>0) 

with land_frac>0 replaced by slmsk==1 if non-fractional orog file. 

if the above test fails, the code should exit and report the error.

ClaraDraper-NOAA:
@GeorgeGayno-NOAA  I have asked @yuanxue2870 to work on this.

GeorgeGayno-NOAA:
What is the typical mismatch? Picking the wrong ocean resolution?

ClaraDraper-NOAA:
> What is the typical mismatch? Picking the wrong ocean resolution?

picking the wrong ocean res, or using the wrong version of the orog files. I tripped over this by trying to use the latest orog to regrid GFSv16 files.

yuanxue2870:
> What is the typical mismatch? Picking the wrong ocean resolution?

Pick the wrong ocean resolution is among one of them. Different fix/orog versions (e.g., 20230615 vs. 20240917) can also cause the problem. Other mismatch reasons may include: fractional vs. non-fractional grid, or e.g., C192 vs. C96 (i.e., different atm. resolution).

PR #1081, opened 2025-07-15
chgres_cube - Add a consistency mask check for restart vs. orog file

yuanxue2870:
## DESCRIPTION OF CHANGES: 
1. Add a consistency check to check the land/sea mask consistency between restart file and orog file
2. Worked with reviewers to better deal with both fractional vs. non-fractional grids for their mask checks

## TESTS CONDUCTED: 

- [x] Compile branch on all Tier 1 machines using Intel. Done using a8d2ad7.
- [x] Compile branch on Ursa using GNU. Done using a8d2ad7.
- [x] Compile branch in 'Debug' mode on WCOSS2. Done using a8d2ad7.
- [x] Compile with Doxygen on any machine with no errors. Done on WCOSS2 using a8d2ad7.
- [x] Run unit tests locally on any Tier 1 machine. Done on WCOSS2 using a8d2ad7. All tests passed.
- [x] Run `chgres_cube` consistency tests locally on all Tier 1 machines. Done using a8d2ad7. All tests passed.

## DEPENDENCIES:
N/A

## DOCUMENTATION:
All new and updated source code must be documented with Doxygen.

- [x] Doxygen is updated.

## ISSUE: 
Fixes issue mentioned in #1078

### DIFF
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1105,7 +1105,7 @@ module sfc_input_data
 
  real(esmf_kind_r8), allocatable :: data_one_tile(:,:) 
  real(esmf_kind_r8), allocatable :: land_frac_one_tile(:,:), vtype_one_tile(:,:)
- real(esmf_kind_i4), allocatable :: slmsk_orog_one_tile(:,:)
+ real(esmf_kind_r4), allocatable :: slmsk_orog_one_tile(:,:)
  real(esmf_kind_r8), allocatable :: data_one_tile_3d(:,:,:)
 
  type(esmf_vm)                   :: vm
@@ -1193,7 +1193,7 @@ module sfc_input_data
      call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
                                    lsoil_input, sfcdata=vtype_one_tile)
      
-     condition1_met = all(((NINT(vtype_one_tile) > 0) .and. (slmsk_orog_one_tile == 1)) .or. ((NINT(vtype_one_tile) <= 0) .and. (slmsk_orog_one_tile == 0)))
+     condition1_met = all(((NINT(vtype_one_tile) > 0) .and. (NINT(slmsk_orog_one_tile) == 1)) .or. ((NINT(vtype_one_tile) <= 0) .and. (NINT(slmsk_orog_one_tile) == 0)))
      condition2_met = all(((land_frac_one_tile > 0.) .and. (NINT(vtype_one_tile) > 0)) .or. ((land_frac_one_tile == 0.) .and. (NINT(vtype_one_tile)<= 0)))
 
      if (condition1_met) then
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1184,7 +1184,7 @@ module sfc_input_data
      ! If this condition not satisfies, restart and orog files are inconsistent
      call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
                                    lsoil_input, sfcdata=vtype_one_tile)
-     if (any((land_frac_one_tile > 0.) .and. (vtype_one_tile <= 0))) then
+     if (any((land_frac_one_tile > 0.) .and. (vtype_one_tile <= 0)) .or. any((land_frac_one_tile == 0.) .and. (vtype_one_tile > 0))) then
        call error_handler("INCONSISTENT LAND/SEA MASK BETWEEN RESTART AND OROG FILES.", 10)
      else
        print*,'mask check ',tile,': consistent!'
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1188,23 +1188,22 @@ module sfc_input_data
      print*,'terrain check ',tile, maxval(data_one_tile)
      error=nf90_close(ncid)
 
-     ! Since the dimensional mismatch case will be captured above
-     ! A consistent fractional grid, will meet both condition1 and condition2
-     ! A consistent non-fractional grid, only can meet condition1 
-     ! An inconsistent fractional(or non-fractional) grid, will not meet condition1
+     ! A consistent non-fractional grid, will meet condition1
+     ! A consistent fractional grid, will meet condition2
      call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
                                    lsoil_input, sfcdata=vtype_one_tile)
      
-     condition1_met = all((vtype_one_tile > 0) .or. (slmsk_orog_one_tile /= 1))
+     condition1_met = all(((vtype_one_tile > 0) .and. (slmsk_orog_one_tile == 1)) .or. ((vtype_one_tile <= 0) .and. (slmsk_orog_one_tile == 0)))
      condition2_met = all(((land_frac_one_tile > 0.) .and. (vtype_one_tile > 0)) .or. ((land_frac_one_tile == 0.) .and. (vtype_one_tile <= 0)))
 
      if (condition1_met) then
-       if (condition2_met) then
+          print*,'[Non-fractional grid]: mask check ',tile,': consistent!'
+     endif
+     if (condition2_met) then
           print*,'[Fractional grid]: mask check ',tile,': consistent!'
-       else
-          print*, '[Non-fractional grid]: mask check ',tile,': consistent!'
-       endif
-     else
+     endif
+
+     if ((.not. condition1_met) .and. (.not. condition2_met)) then
        call error_handler("INCONSISTENT LAND/SEA MASK BETWEEN RESTART AND OROG FILES.", 10)
      endif
 
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1193,8 +1193,8 @@ module sfc_input_data
      call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
                                    lsoil_input, sfcdata=vtype_one_tile)
      
-     condition1_met = all(((vtype_one_tile > 0) .and. (slmsk_orog_one_tile == 1)) .or. ((vtype_one_tile <= 0) .and. (slmsk_orog_one_tile == 0)))
-     condition2_met = all(((land_frac_one_tile > 0.) .and. (vtype_one_tile > 0)) .or. ((land_frac_one_tile == 0.) .and. (vtype_one_tile <= 0)))
+     condition1_met = all(((NINT(vtype_one_tile) > 0) .and. (slmsk_orog_one_tile == 1)) .or. ((NINT(vtype_one_tile) <= 0) .and. (slmsk_orog_one_tile == 0)))
+     condition2_met = all(((land_frac_one_tile > 0.) .and. (NINT(vtype_one_tile) > 0)) .or. ((land_frac_one_tile == 0.) .and. (NINT(vtype_one_tile)<= 0)))
 
      if (condition1_met) then
           print*,'[Non-fractional grid]: mask check ',tile,': consistent!'
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1437,6 +1437,7 @@ module sfc_input_data
  enddo TILE_LOOP
 
  deallocate(data_one_tile, data_one_tile_3d)
+ deallocate(vtype_one_tile, land_frac_one_tile)
 
  end subroutine read_input_sfc_restart_file
 
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1101,17 +1101,18 @@ module sfc_input_data
 
  integer                         :: error, rc
  integer                         :: id_dim, idim_input, jdim_input
- integer                         :: ncid, tile, id_var, idum(2)
+ integer                         :: ncid, tile, id_var, id_var1, idum(2)
 
- real(esmf_kind_r8), allocatable :: data_one_tile(:,:)
+ real(esmf_kind_r8), allocatable :: data_one_tile(:,:), land_frac_one_tile(:,:), vtype_one_tile(:,:)
  real(esmf_kind_r8), allocatable :: data_one_tile_3d(:,:,:)
 
  type(esmf_vm)                   :: vm
 
-!---------------------------------------------------------------------------
+!------------------------------------------------------------------------------------------
 ! Get i/j dimensions and number of soil layers from first surface file.
 ! Do dimensions match those from the orography file?
-!---------------------------------------------------------------------------
+! Do land/sea masks (i.e., ocean res. and orog version) match that from the orography file?
+!------------------------------------------------------------------------------------------
 
  print*,"- CALL VMGetGlobal"
  call ESMF_VMGetGlobal(vm, rc=rc)
@@ -1151,9 +1152,13 @@ module sfc_input_data
 
  if (localpet == 0) then
    allocate(data_one_tile(idim_input,jdim_input))
+   allocate(land_frac_one_tile(idim_input,jdim_input))
+   allocate(vtype_one_tile(idim_input,jdim_input))
    allocate(data_one_tile_3d(idim_input,jdim_input,lsoil_input))
  else
    allocate(data_one_tile(0,0))
+   allocate(land_frac_one_tile(0,0))
+   allocate(vtype_one_tile(0,0))
    allocate(data_one_tile_3d(0,0,0))
  endif
 
@@ -1166,10 +1171,25 @@ module sfc_input_data
      call netcdf_err(error, 'OPENING OROGRAPHY FILE' )
      error=nf90_inq_varid(ncid, 'orog_raw', id_var)
      call netcdf_err(error, 'READING OROG RECORD ID' )
+     error=nf90_inq_varid(ncid, 'land_frac', id_var1)
+     call netcdf_err(error, 'READING OROG LAND_FRAC ID' )
      error=nf90_get_var(ncid, id_var, data_one_tile)
      call netcdf_err(error, 'READING OROG RECORD' )
+     error=nf90_get_var(ncid, id_var1, land_frac_one_tile)
+     call netcdf_err(error, 'READING OROG LAND_FRAC' )
      print*,'terrain check ',tile, maxval(data_one_tile)
      error=nf90_close(ncid)
+
+     ! Check if land_frac > 0, vtype > 0
+     ! If this condition not satisfies, restart and orog files are inconsistent
+     call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
+                                   lsoil_input, sfcdata=vtype_one_tile)
+     if (any((land_frac_one_tile > 0.) .and. (vtype_one_tile <= 0))) then
+       call error_handler("INCONSISTENT LAND/SEA MASK BETWEEN RESTART AND OROG FILES.", 10)
+     else
+       print*,'mask check ',tile,': consistent!'
+     endif
+
    endif
 
    print*,"- CALL FieldScatter FOR INPUT TERRAIN."
--- a/sorc/chgres_cube.fd/sfc_input_data.F90
+++ b/sorc/chgres_cube.fd/sfc_input_data.F90
@@ -1101,17 +1101,19 @@ module sfc_input_data
 
  integer                         :: error, rc
  integer                         :: id_dim, idim_input, jdim_input
- integer                         :: ncid, tile, id_var, id_var1, idum(2)
+ integer                         :: ncid, tile, id_var, id_var1, id_var2, idum(2)
 
- real(esmf_kind_r8), allocatable :: data_one_tile(:,:), land_frac_one_tile(:,:), vtype_one_tile(:,:)
+ real(esmf_kind_r8), allocatable :: data_one_tile(:,:) 
+ real(esmf_kind_r8), allocatable :: land_frac_one_tile(:,:), vtype_one_tile(:,:)
+ real(esmf_kind_i4), allocatable :: slmsk_orog_one_tile(:,:)
  real(esmf_kind_r8), allocatable :: data_one_tile_3d(:,:,:)
 
  type(esmf_vm)                   :: vm
-
+ logical                         :: condition1_met, condition2_met
 !------------------------------------------------------------------------------------------
 ! Get i/j dimensions and number of soil layers from first surface file.
 ! Do dimensions match those from the orography file?
-! Do land/sea masks (i.e., ocean res. and orog version) match that from the orography file?
+! Do land/sea masks (i.e., ocean res. and orog version) match those from the orography file?
 !------------------------------------------------------------------------------------------
 
  print*,"- CALL VMGetGlobal"
@@ -1153,11 +1155,13 @@ module sfc_input_data
  if (localpet == 0) then
    allocate(data_one_tile(idim_input,jdim_input))
    allocate(land_frac_one_tile(idim_input,jdim_input))
+   allocate(slmsk_orog_one_tile(idim_input,jdim_input))
    allocate(vtype_one_tile(idim_input,jdim_input))
    allocate(data_one_tile_3d(idim_input,jdim_input,lsoil_input))
  else
    allocate(data_one_tile(0,0))
    allocate(land_frac_one_tile(0,0))
+   allocate(slmsk_orog_one_tile(0,0))
    allocate(vtype_one_tile(0,0))
    allocate(data_one_tile_3d(0,0,0))
  endif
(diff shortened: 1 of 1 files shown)


BrianCurtis-NOAA:
WCOSS2 Debug build went OK with 63a2121

yuanxue2870:
Please review: @ClaraDraper-NOAA, @GeorgeGayno-NOAA. Thank you!

sorc/chgres_cube.fd/sfc_input_data.F90
```diff
@@ -1166,10 +1171,25 @@ subroutine read_input_sfc_restart_file(localpet)
...
+     error=nf90_get_var(ncid, id_var1, land_frac_one_tile)
+     call netcdf_err(error, 'READING OROG LAND_FRAC' )
      print*,'terrain check ',tile, maxval(data_one_tile)
      error=nf90_close(ncid)
+
+     ! Check if land_frac > 0, vtype > 0
+     ! If this condition not satisfies, restart and orog files are inconsistent
+     call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
+                                   lsoil_input, sfcdata=vtype_one_tile)
+     if (any((land_frac_one_tile > 0.) .and. (vtype_one_tile <= 0))) then
```
ClaraDraper-NOAA: we also need to check for the opposite error: 
land_frac==0 and vtype>0

ClaraDraper-NOAA (changes requested):
Thanks @yuanxue2870  There's one small change needed.

ClaraDraper-NOAA:
Above, you mention that the C384 test is inconsistent - so we'll need to update that test before this can be merged. It may just be a namelist issue - is there a C384 orog file that will work? (i.e., does the namelist just have the wrong ocean res?)

sorc/chgres_cube.fd/sfc_input_data.F90
yuanxue2870: OK. Will add an extra if statement here. Convert to draft while editing…

yuanxue2870:
> 2\. /scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc

> Above, you mention that the C384 test is inconsistent - so we'll need to update that test before this can be merged. It may just be a namelist issue - is there a C384 orog file that will work? (i.e., does the namelist just have the wrong ocean res?)

Good question. I do have a consistent C384 (fractional grid) test that can be used to swap with the current reg-test once we are settled on the code part.

I will show you in a bit more details tomorrow during our meeting -- hard to tell the ocean resolution from the restart file since I do not know who creates it. On the other hand, the orog files `/scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc` are identical to those in  
`/scratch3/NCEPDEV/global/role.glopara/fix/orog/20230615/C384/C384_oro_data.tile[xx].nc`, which seems to be an outdated non-fractional C384 grid (please feel free to correct me if I am wrong, I suspect it might be non-fractional grid based on the directory structure). So in this case, I personally think it would be better to replace with consistent new restart with the latest orog files.

ClaraDraper-NOAA:
> > 2. /scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc
> 
> > Above, you mention that the C384 test is inconsistent - so we'll need to update that test before this can be merged. It may just be a namelist issue - is there a C384 orog file that will work? (i.e., does the namelist just have the wrong ocean res?)
> 
> Good question. I do have a consistent C384 (fractional grid) test that can be used to swap with the current reg-test once we are settled on the code part.
> 
> I will show you in a bit more details tomorrow during our meeting -- hard to tell the ocean resolution from the restart file since I do not know who creates it. The orog files `/scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc` are identical to those in `/scratch3/NCEPDEV/global/role.glopara/fix/orog/20230615/C384/C384_oro_data.tile[xx].nc`, which seems to be an outdated non-fractional C384 grid. So in this case, I personally think it would be better to replace with consistent new restart with the latest orog files.

Let's let @GeorgeGayno-NOAA  weigh in on this - I believe that he intentionally archives the orography files that are used for the tests so that the tests don't all fail when the default orography files are changed. So it may be better to update the restarts instead.

yuanxue2870:
> > > 2. /scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc
> > 
> > 
> > > Above, you mention that the C384 test is inconsistent - so we'll need to update that test before this can be merged. It may just be a namelist issue - is there a C384 orog file that will work? (i.e., does the namelist just have the wrong ocean res?)
> > 
> > 
> > Good question. I do have a consistent C384 (fractional grid) test that can be used to swap with the current reg-test once we are settled on the code part.
> > I will show you in a bit more details tomorrow during our meeting -- hard to tell the ocean resolution from the restart file since I do not know who creates it. The orog files `/scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc` are identical to those in `/scratch3/NCEPDEV/global/role.glopara/fix/orog/20230615/C384/C384_oro_data.tile[xx].nc`, which seems to be an outdated non-fractional C384 grid. So in this case, I personally think it would be better to replace with consistent new restart with the latest orog files.
> 
> Let's let @GeorgeGayno-NOAA weigh in on this - I believe that he intentionally archives the orography files that are used for the tests so that the tests don't all fail when the default orography files are changed. So it may be better to update the restarts instead.

Definitely! Just a minor add-on that it seems to me these C384 input/orog files are only used for the reg-test of `c96.fv3.restart`, so once I swap with new restart and new orog file, I did run all other regression tests as well as all unit tests, they are all passed.

sorc/chgres_cube.fd/sfc_input_data.F90
yuanxue2870: Done.

ClaraDraper-NOAA: Thanks Yuan. I'll approve this, assuming George can advise on the tests.

ClaraDraper-NOAA (approved):

sorc/chgres_cube.fd/sfc_input_data.F90
```diff
@@ -1151,9 +1152,13 @@ subroutine read_input_sfc_restart_file(localpet)
 
  if (localpet == 0) then
    allocate(data_one_tile(idim_input,jdim_input))
+   allocate(land_frac_one_tile(idim_input,jdim_input))
+   allocate(vtype_one_tile(idim_input,jdim_input))
```
GeorgeGayno-NOAA, line 1159: Please deallocate these arrays after the 'if' block.

GeorgeGayno-NOAA:
> > > 2. /scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc
> > 
> > 
> > > Above, you mention that the C384 test is inconsistent - so we'll need to update that test before this can be merged. It may just be a namelist issue - is there a C384 orog file that will work? (i.e., does the namelist just have the wrong ocean res?)
> > 
> > 
> > Good question. I do have a consistent C384 (fractional grid) test that can be used to swap with the current reg-test once we are settled on the code part.
> > I will show you in a bit more details tomorrow during our meeting -- hard to tell the ocean resolution from the restart file since I do not know who creates it. The orog files `/scratch3/NCEPDEV/nems/role.ufsutils/ufs_utils/reg_tests/chgres_cube/fix/C384/C384_oro_data.tile[xx].nc` are identical to those in `/scratch3/NCEPDEV/global/role.glopara/fix/orog/20230615/C384/C384_oro_data.tile[xx].nc`, which seems to be an outdated non-fractional C384 grid. So in this case, I personally think it would be better to replace with consistent new restart with the latest orog files.
> 
> Let's let @GeorgeGayno-NOAA weigh in on this - I believe that he intentionally archives the orography files that are used for the tests so that the tests don't all fail when the default orography files are changed. So it may be better to update the restarts instead.

Yes, all input data for the consistency tests are stored in directory owned by the role account, so any updates to the 'official' fixed files does not cause the test to fail.

I think the test data is ok. The restart files were created before the fractional grid. I visualized the tile 1 restart. The vegetation type data is clearly not defined for a fractional grid. So, you can't expect it to match the land fraction array. At some point, we stop supporting older data. If we adopt your logic, we won't be able to use pre-fractional grid restart files.

sorc/chgres_cube.fd/sfc_input_data.F90
GeorgeGayno-NOAA: Thanks!

sorc/chgres_cube.fd/sfc_input_data.F90
```diff
@@ -1166,10 +1171,25 @@ subroutine read_input_sfc_restart_file(localpet)
...
      call netcdf_err(error, 'READING OROG RECORD' )
+     error=nf90_get_var(ncid, id_var1, land_frac_one_tile)
+     call netcdf_err(error, 'READING OROG LAND_FRAC' )
      print*,'terrain check ',tile, maxval(data_one_tile)
      error=nf90_close(ncid)
+
+     ! Check if land_frac > 0, vtype > 0
+     ! If this condition not satisfies, restart and orog files are inconsistent
+     call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
+                                   lsoil_input, sfcdata=vtype_one_tile)
```
GeorgeGayno-NOAA, line 1194: In the issue (#1078) description, you mention checking against 'slmsk' for non-fractional grids. Do you want to include that check? Are there any attributes in the restart files that indicate a fractional grid?

yuanxue2870: @ClaraDraper-NOAA wrote the issue description, so please wait for her to weigh in. I personally try to stay away from slmsk check for this case in that "slmsk" (with the same variable name) has different definitions/values in orog vs. restart file. I personally find using "land_frac > 0, vtype > 0" check has been the most straightforward check for now for fractional grid. I may have not looked into the non-fractional grid and their restarts that much, but should not "land_frac > 0, vtype > 0" logic applies as well?

ClaraDraper-NOAA: @GeorgeGayno-NOAA do we need backwards compatibility to make sure this code works with non-fractional grid cells?

If we do, then this code will need to work with non-fractional orog files (which have slmsk instead of land_frac). We can use slmsk in that case, as it's well-defined for non-fractional grid cells.

yuanxue2870: > @GeorgeGayno-NOAA do we need backwards compatibility to make sure this code works with non-fractional grid cells?
> 
> If we do, then this code will need to work with non-fractional orog files (which have slmsk instead of land_frac). We can use slmsk in that case, as it's well-defined for non-fractional grid cells.

Hmm, this is actually my question here: /scratch3/NCEPDEV/global/role.glopara/fix/orog/20230615/C384/C384_oro_data.tile[xx].nc are presumably being non-fractional grid. But they do have land_frac field in the orog file.

GeorgeGayno-NOAA: > @GeorgeGayno-NOAA do we need backwards compatibility to make sure this code works with non-fractional grid cells?
> 
> If we do, then this code will need to work with non-fractional orog files (which have slmsk instead of land_frac). We can use slmsk in that case, as it's well-defined for non-fractional grid cells.

If it is straightforward to add logic to handle non-fractional grids, then I recommend we do so. The `gdas_init` utility still supports GFS v15 input data (which is restart format). However, at some point we stop supporting older data. So, if it is not easy to add the additional logic, then we can drop it.

ClaraDraper-NOAA: To answer my own question: we do need backwards compatibility, so that this can be applied to GFSv16 restarts. My understanding is that they were generated with orog files that do not have land_frac in them.

ClaraDraper-NOAA: > > @GeorgeGayno-NOAA do we need backwards compatibility to make sure this code works with non-fractional grid cells?
> > If we do, then this code will need to work with non-fractional orog files (which have slmsk instead of land_frac). We can use slmsk in that case, as it's well-defined for non-fractional grid cells.
> 
> If it is straightforward to add logic to handle non-fractional grids, then I recommend we do so. The `gdas_init` utility still supports GFS v15 input data (which is restart format). However, at some point we stop supporting older data. So, if it is not easy to add the additional logic, then we can drop it.

I think we could just add logic that: 
1. Looks for land_frac
2. If it can't find it, switch to using slmsk.

yuanxue2870: Thanks for all your input. The land_frac ==> slmsk check mentioned above will not work in that all org files (from version 20230615 to 20240917), be it fractional grid or non-fractional grid, have land_frac, so the slmsk check will never be conducted. Convert the PR to draft when I am plumbing through other logic to better incorporate land/sea mask check for non-fractional grid. Thanks for your patience.

GeorgeGayno-NOAA: Can you check veg type against land_frac AND slmask? If one of the checks passes, you have a correct orog file, no?

yuanxue2870: Thanks for your suggestion. Please see the code edits. The original reg-test is in fact a consistent non-fractional grid test case. So no new test data is needed. Plus, both consistent and inconsistent fractional grid tests were done offline to make sure the code is working properly. Thanks for re-reviewing.

sorc/chgres_cube.fd/sfc_input_data.F90
```diff
@@ -1173,21 +1177,35 @@ subroutine read_input_sfc_restart_file(localpet)
...
      call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
                                    lsoil_input, sfcdata=vtype_one_tile)
-     if (any((land_frac_one_tile > 0.) .and. (vtype_one_tile <= 0)) .or. any((land_frac_one_tile == 0.) .and. (vtype_one_tile > 0))) then
-       call error_handler("INCONSISTENT LAND/SEA MASK BETWEEN RESTART AND OROG FILES.", 10)
+     
+     condition1_met = all((vtype_one_tile > 0) .or. (slmsk_orog_one_tile /= 1))
+     condition2_met = all(((land_frac_one_tile > 0.) .and. (vtype_one_tile > 0)) .or. ((land_frac_one_tile == 0.) .and. (vtype_one_tile <= 0)))
+
+     if (condition1_met) then
+       if (condition2_met) then
```
BrianCurtis-NOAA: If there is a condition2 to be met where condition1 is not, then this will not work. Is that the goal here?

yuanxue2870: > If there is a condition2 to be met where condition1 is not, then this will not work. Is that the goal here?

Yes. Based on my own analysis, condition1 must be met in both frac and non-frac cases. Where condition1 is not met, there is no need to analyze condition2.

GeorgeGayno-NOAA: @yuanxue2870 - For non-fractional cases, slmask should match veg type. It won't match land frac. For fractional cases, land frac should match veg type. It won't match slmask. So, I think the logic should check if either condition 1 or condition 2 is true. If they are both false, then abort.

GeorgeGayno-NOAA: @yuanxue2870 - what I propose should be easy to verify. Do you have examples of both cases?

yuanxue2870: > @yuanxue2870 - For non-fractional cases, slmask should match veg type. It won't match land frac. For fractional cases, land frac should match veg type. It won't match slmask. So, I think the logic should check if either condition 1 or condition 2 is true. If they are both false, then abort.

Your logic here is also true - I am pretty sure it is going to work and I am happy to revise as suggested. Yes, I have test cases. But one minor concern is that your proposed logic will not be able to distinguish between fractional grid vs. non_fractional grid mask consistency check as shown in the code here because in the fractional case, condition1 is also true (i.e., where slmsk==1, vtype > 0). If this does not matter, please let me know. I will revise the code accordingly based on your logic. Thanks!

GeorgeGayno-NOAA: @yuanxue2870 - Let me look at your fractional and non-fractional test cases.

BrianCurtis-NOAA:
Started on  ufe02
Commit hash: 24743de3fe018a337e4946f231ee58c1d401cd68

regrid_sfc consistency tests PASSED
weight_gen consistency tests PASSED
ocnice_prep consistency tests PASSED
cpld_gridgen consistency tests PASSED
chgres_cube consistency tests PASSED
grid_gen consistency tests PASSED
global_cycle consistency tests PASSED
ice_blend consistency tests PASSED
snow2mdl consistency tests PASSED

sorc/chgres_cube.fd/sfc_input_data.F90
yuanxue2870: Sure. On Ursa, I have three test cases along with log files. Please go to the directory: /scratch4/NCEPDEV/stmp/Yuan.Xue/pull_request/UFS_UTILS/reg_tests/chgres_cube

test case 1: c96.fv3.restart.sh_ori (Original, good non-frac case) ==> goodnonfrac_consistency.log01
test case 2: c96.fv3.restart.sh_good_frac (C384mx025, good frac case) ==> goodfrac_consistency.log01
test case 3: c96.fv3.restart.sh_bad_frac (intentionally mix orog, bad frac case) ==> badfrac_consistency.log01

Let me know if there are any questions. Thanks!

sorc/chgres_cube.fd/sfc_input_data.F90
GeorgeGayno-NOAA: Can you rerun your tests and print the value of both `condition1_met` and `condition2_met`? I am curious what it shows.

yuanxue2870: > Can you rerun your tests and print the value of both `condition1_met` and `condition2_met`? I am curious what it shows.

Sure. I have re-run these tests on Ursa and print both logical variables to the log. Please go to the directory: /scratch4/NCEPDEV/stmp/Yuan.Xue/pull_request/UFS_UTILS/reg_tests/chgres_cube

test case 1: c96.fv3.restart.sh_ori (Original, good non-frac case) ==> goodnonfrac_consistency.log ==> condition1_met=T, condition2_met=F
test case 2: c96.fv3.restart.sh_good_frac (C384mx025, good frac case) ==> goodfrac_consistency.log ==> condition1_met=T, condition2_met=T
test case 3: c96.fv3.restart.sh_bad_frac (intentionally mix orog, bad frac case) ==> badfrac_consistency.log==> condition1_met=F, condition2_met=F

Let me know if there are any questions. Thanks!

GeorgeGayno-NOAA: Thanks. I am not sure I understand the logic for 'condition1'. For test two, which is fractional, veg type will not match slmask. Veg type will be non-zero for all points with a land fraction > 0. slmask is 0/1 based on a 50% land fraction.

Will you be in the office tomorrow? It would be easier to discuss in person.

yuanxue2870: Sure! But I will not be in the office tomorrow. Will you be in the office on Thursday? Condition1 is essentially to say: whenever slmsk==1, vtype > 0.

ClaraDraper-NOAA:
I'm also not sure about the logic on condition 1.

For condition 2, I want to clarify to things, as I gave bad advice above: 
1. The new code should work with GFSv16 restarts (not fractional grids) 
2. I thought that the restarts had either land_frac (for fractional) or slmsk (non-fractional). This is wrong - all restarts have both. @yuanxue2870 If you look at a GFSv16 example restart you should be able to figure out what the relationship is between the land_frac, slmsk and veg_type and design a test for that.

yuanxue2870:
> I'm also not sure about the logic on condition 1.
> 
> For condition 2, I want to clarify to things, as I gave bad advice above:
> 
> 1. The new code should work with GFSv16 restarts (not fractional grids)
> 2. I thought that the restarts had either land_frac (for fractional) or slmsk (non-fractional). This is wrong - all restarts have both. @yuanxue2870 If you look at a GFSv16 example restart you should be able to figure out what the relationship is between the land_frac, slmsk and veg_type and design a test for that.

Thanks. These two logics are in fact based on my own analysis of land_frac, slmsk, and vtype for both consistent frac and non-frac cases. Condition 1 is equivalent to whenever slmsk==1, vtype >0; logic 2 is equivalent to when land_frac>(=)0, vtype>(=)0. Condition 1 looks not as straightforward as Condition 2 (i.e., it did look tricky!) because for frac grid, there can be cases where vtype>0, slmsk=0. So Logic 1 has to be very carefully formulated to make sure it is identical to whenever slmsk==1, vtype >0. I also consulted with our orog expert Sanath to make sure the analyzed slmsk and vtype relation in Logic1 is true for both grids. Test cases of both frac and non-frac grids are also approving this logic set. I have a meeting with George this afternoon to discuss in details. Let me know if there are any questions.

sorc/chgres_cube.fd/sfc_input_data.F90
CoryMartin-NOAA: Sorry to pile in on this, but I also question the condition 1 case. In general, "or" can be dangerous to use. This isn't saying that these are exclusive, what if there is a point where vtype is greater than 0 and the slmsk is not 1?

yuanxue2870: Let's think of the reverse statement of condtion1, so the reverse statement of condition1 is: "any((vtype_one_tile <= 0) .and. (slmsk_orog_one_tile == 1))". This means condition1 is equivalent to that **no** instance should exist where vtype<=0 and slmsk==1, if there exists a case, exist the code. If there is a point where vtype is greater than 0 and the slmsk is not 1, condition1_met is still true, and it is gonna check condition2_met to see if it is a consistent fractional grid case. I just finished discussing with George, I will need to talk with you and Clara regarding the discussions. Thanks!

CoryMartin-NOAA: ok, perhaps I need to double check how Fortran handles these 'or' exceptions, because often languages will not continue beyond the first 'true'. So if vtype is greater than zero, it will just be happy and move on regardless of if slmsk at that same point is correct.

sorc/chgres_cube.fd/sfc_input_data.F90
```diff
@@ -1188,23 +1188,22 @@ subroutine read_input_sfc_restart_file(localpet)
...
-     ! A consistent fractional grid, will meet both condition1 and condition2
-     ! A consistent non-fractional grid, only can meet condition1 
-     ! An inconsistent fractional(or non-fractional) grid, will not meet condition1
+     ! A consistent non-fractional grid, will meet condition1
+     ! A consistent fractional grid, will meet condition2
      call read_fv3_grid_data_netcdf('vtype', tile, idim_input, jdim_input, &
                                    lsoil_input, sfcdata=vtype_one_tile)
      
-     condition1_met = all((vtype_one_tile > 0) .or. (slmsk_orog_one_tile /= 1))
+     condition1_met = all(((vtype_one_tile > 0) .and. (slmsk_orog_one_tile == 1)) .or. ((vtype_one_tile <= 0) .and. (slmsk_orog_one_tile == 0)))
```
CoryMartin-NOAA: This updated logic looks better to me

CoryMartin-NOAA: @CatherineThomas-NOAA FYI since we need this for v17

GeorgeGayno-NOAA: Variable `slmsk_orog_one_tile` is floating point. I might add a NINT here. Same with `vtype_one_tile`. Both variables are categorical.

yuanxue2870: > Variable `slmsk_orog_one_tile` is floating point. I might add a NINT here. Same with `vtype_one_tile`. Both variables are categorical.

Thanks George for your comment. I set "real(esmf_kind_**i4**), allocatable :: slmsk_orog_one_tile(:,:)" at Line 1108, will that still need an "NINT" around it?

GeorgeGayno-NOAA: "esmf_kind_i4" is the ESMF size of a four byte integer. You are still declaring "slmsk_orog_one_tile" as 'real'.  When declaring as 'real' you should use 'esmf_kind_**r4**' for four byte floats or 'esmf_kind_**r8**' for double precision. Since 'slmsk' in the 'orog' file is four byte real, I would use 'real(esmf_kind_r4), allocatable :: slmsk_orog_one_tile(:,:)', then do a NINT.

yuanxue2870: Thanks for your explanation! Done.

GeorgeGayno-NOAA (approved):

GeorgeGayno-NOAA:
@yuanxue2870 - 'develop' was just updated. Please merge the updates to your branch, then I will merge this PR. The updates don't affect `chgres_cube`.

yuanxue2870:
> @yuanxue2870 - 'develop' was just updated. Please merge the updates to your branch, then I will merge this PR. The updates don't affect `chgres_cube`.

Done. Thanks!

BrianCurtis-NOAA (approved):

